PR #7, opened 2026-02-27
Add Toolpath class, example 15, and fix merge_animation_points for duplicate timestamps

tdamsma:
## Summary

- **`Toolpath` class** (`src/threejs_viewer/toolpath.py`): wraps G-code-style or continuous-path data in `(N,6) [t,x,y,z,width,height]` format
  - `Toolpath.from_gcode(raw, bead_width, bead_height)` — detects extrusion from `dE > 0`, computes time from feedrate
  - `Toolpath.from_points(points, w, h, duration)` — continuous paths (spiral vases, trajectories)
  - `.frame_times(n)`, `.merge(frame_times)` — segment-aligned animation helpers
  - `.colorize(color_or_values, colormap, travel_color)` — stores per-point colors; dispatch by type: string (arc-length gradient), int (hex), tuple/`(3,)` array (solid RGB), `(N,)` array (per-point scalar mapped through colormap)
  - `.colors` property + setter
  - `.to_mesh()` — builds `{positions, indices, normals, colors}` dict; use with `v.add_mesh("id", **tp.to_mesh())`

- **Example 15** (`examples/15_toolpath_interrupted.py`): pill/racetrack toolpath with interrupted extrusion (travel moves), G-code-style format, plasma colormap, `merge_animation_points` for clean segment-aligned draw_range animation

- **Example 11** updated to use `Toolpath.from_points` + `colorize("viridis")` + `add_mesh(**tp.to_mesh())`

- **Bug fix in `Toolpath.from_gcode`**: arc→straight→arc join points share the same xyz, so the arriving segment has length 0 → `dE=0` → those points were incorrectly assigned `w=0`, creating spurious cap rings inside the extrusion section. Fix: when the arriving segment is zero-length but the departing segment is extruding, promote to full width. Index 0 is excluded to preserve the start cap.

- **Bug fix in `merge_animation_points`**: a frame_time just after an extrusion→travel transition was inheriting `w=bw` instead of `w=0` because `np.interp` uses the leftmost duplicate timestamp as its anchor. Fixed by using `searchsorted(side='right') - 1` for interpolation bracketing.

- **Removed `add_bead()` from `ViewerClient`** — geometry now lives entirely in `Toolpath.to_mesh()`; the call site is simply `v.add_mesh("id", **tp.to_mesh())`.

## Test plan

- [x] `uv run pytest -p no:playwright` — all 78 tests pass
- [x] `uv run python examples/11_toolpath.py` — spiral vase with viridis gradient, no visual artefacts
- [x] `uv run python examples/15_toolpath_interrupted.py` — pill toolpath, travel sections invisible, extrusion tapers cleanly with no pinches at arc/straight joins


tdamsma:
Superseded by a new PR — branch has evolved significantly since this was opened.

PR #20, opened 2026-04-09
Add grid API, fix animation cleanup, extend query_scene

tdamsma:
## Summary
- **Grid API**: `show_grid(visible, size, divisions)` — toggles/resizes the ground grid (hidden by default). Properly disposes old Three.js geometry on resize.
- **Animation cleanup**: `stop_animation()` resets draw ranges to 1.0 so bead meshes don't stay partially drawn. `clear()` now resets animation state so stale animations don't reference deleted objects.
- **query_scene extensions**: Response now includes per-object `drawRange`, `_animation.playing`, and `_grid.visible` — enables proper browser integration testing.
- **3 new browser tests**: draw range reset on stop, clear resets animation state, grid toggle.

## Motivation
ribweaver's panel slicer has a step-based workflow where going back from animation (step 4) to static toolpath (step 3) left stale animation state — the bead stayed partially drawn and animation controls remained visible. There was no clean way to reset to static display. The grid was also only removable by reaching into private `_gridHelper`/`_scene` internals.

## Test plan
- [x] All 123 tests pass (120 unit + 3 new browser integration)
- [x] Linting clean (`ruff check` + `ruff format --check`)
- [ ] Manual test: load animation with draw_ranges, stop it, verify full mesh visible
- [ ] Manual test: `viewer.show_grid(True, 20, 20)` then `viewer.show_grid(False)`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #21, opened 2026-04-09
Fix async race conditions and optimize animation channel refs

tdamsma:
## Summary
- **Refs cache for draw_ranges/clip_times**: These two channel apply functions were the only ones doing `Map.get()` lookups every animation frame instead of using the `refs` cache. Now consistent with transforms/colors/visibility/opacity handlers.
- **Generation counters**: Added `_sceneGeneration` (incremented on `_clearScene` and reconnect) and `_animGeneration` (incremented on `_loadAnimation` and reconnect). All 4 async HTTP fetch handlers (`load_animation_http`, `add_model_binary`, `add_polyline_binary`, `add_mesh_binary`) capture the generation before `await fetch()` and discard the result if it changed — prevents stale fetches from corrupting cleared/replaced state.
- **`_pending_fetches` in query_scene**: Backend can now check `tree["_pending_fetches"]` to know if objects are still loading, without needing `wait_for_assets()`.

## Motivation
Audit found that `clear_scene` during in-flight HTTP fetches silently added objects to a cleared scene, and two rapid `load_animation_http` calls could race (slower fetch overwrites newer animation). The refs cache inconsistency was flagged by TS diagnostics.

## Test plan
- [x] All 120 tests pass
- [x] Linting clean
- [ ] Manual: send `clear_scene` while a large model is loading, verify it doesn't appear
- [ ] Manual: rapidly switch animations, verify correct one plays

### DIFF
--- a/src/threejs_viewer/client.py
+++ b/src/threejs_viewer/client.py
@@ -1072,7 +1072,10 @@ class ViewerClient:
     def query_scene(self, timeout: float = 5.0) -> dict:
         """Query the viewer's scene graph.
 
-        Returns dict of {id: {type, parent, children, visible}}.
+        Returns dict with keys:
+
+        - Object IDs mapping to ``{type, parent, children, visible}``
+        - ``"_pending_fetches"`` — number of in-flight HTTP fetches (int)
         """
         request_id = str(uuid.uuid4())
         event = threading.Event()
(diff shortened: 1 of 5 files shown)


tdamsma:
Combining into a single PR with #20 and Copilot review feedback.

PR #36, opened 2026-04-22
Per-tube LOD configuration for parametric tubes

tdamsma:
## Summary
- Add `lod=` kwarg to `add_parametric_tube` (transparently passed through by `add_toolpath`) so callers can disable LOD entirely (`lod=False`) or tune `epsilon_divisor` / `threshold` per tube — previously both were hardcoded globals (`LOD_EPSILON_DIVISOR = 2500`, `n >= 25000`).
- Thread the config through to the LOD worker's `init` message so the ongoing 2 Hz updates respect the override, not just the initial synchronous reduction.
- Defaults are unchanged: omitting `lod` preserves the prior 2500/25000 behavior, so no existing caller needs to update.

### API shape
```python
v.add_parametric_tube("bead", spine, widths, heights, lod=False)                      # disable entirely
v.add_parametric_tube("hires", spine, widths, heights, lod={"epsilon_divisor": 10000}) # keep more detail
v.add_parametric_tube("small", spine, widths, heights, lod={"threshold": 0})          # force LOD on below 25k
```

Invalid shapes (`lod=True`, unknown keys, non-positive `epsilon_divisor`, negative `threshold`, non-dict) raise `ValueError` client-side before the message is sent.

### Motivation
Bead-inspection workflows (Ribweaver) need every original spine point preserved; the default simplification collapses visible detail on large toolpaths. Rather than bump a global constant (which imposes one consumer's tuning on every other caller), the knob moves to the call site. The wire format uses camelCase (`{"epsilonDivisor": ..., "threshold": ...}`) consistent with existing tube keys (`numSpinePoints`, `upVector`, `heightOffset`).

## Test plan
- [x] `uv run pytest` — 180 tests pass, including the 9 new `lod` serialization/validation cases and 1 `add_toolpath` passthrough test
- [x] `uv run ruff check . && uv run ruff format --check .` — clean
- [x] `node --check src/threejs_viewer/viewer/viewer.js` — no syntax errors
- [x] `npx tsc --noEmit -p jsconfig.json` — JSDoc types clean
- [x] Browser integration tests (Playwright, chromium) for parametric tube primitive still pass
- [ ] Manual smoke: spin up `examples/11_toolpath.py` with `lod=False` on one tube and `lod={"epsilon_divisor": 10000}` on another, eyeball in viewer


tdamsma:
Both Copilot comments addressed in 2d74f35:

- **`threshold` silent float truncation** → now rejected with `ValueError` via `isinstance(thr, numbers.Integral)`. Float thresholds (`1.9`) no longer silently become `1`.
- **numpy scalars rejected** → switched to `numbers.Real` / `numbers.Integral` ABCs, so `np.float32`, `np.int64`, etc. are accepted (consistent with the rest of the client API). `bool` is still explicitly rejected.

Added two tests: `test_add_parametric_tube_lod_threshold_float_rejected` and `test_add_parametric_tube_lod_accepts_numpy_scalars`.

PR #41, opened 2026-04-22
Fix twisted parametric tube at 180 deg tangent reversals

tdamsma:
## Summary
- Parametric-tube spines that hairpin back on themselves (retrace, zigzag infill) produced a self-intersecting figure-8 twist at the reversal. Root cause: the constant-up frame derives `U = V × T`, so when `T` flips sign `U` flips too — adjacent rings then index their cross-section vertices at opposite physical points.
- Fix: after the main frame loop, sweep `localFrames` and — where `U_i · U_{i-1} < -0.95` — negate `U` on that ring and rewrite its vertices. `V` is untouched (even in `T`), so no cascading changes. Sweep is sticky: the whole return leg stays aligned with the forward leg. Mirrored in the LOD worker so reduced and full-resolution geometries agree at reversals. Skipped entirely when the caller supplies explicit `orientations`.
- Threshold `-0.95` restricts the flip to true hairpins; gradual curves and non-reversal sharp corners (<160°) are untouched.

## Perf
Measured on V8: detection is ~3 ms for a 1M-point spine (one dot product + compare per ring, sequential through localFrames while still hot in L1). The rewrite fires only at actual hairpins, so practical cost is ~zero for all realistic tubes. Build-time only — no runtime frame-render impact.

## Demo / regression test
`examples/20_toolpath_zigzag.py` draws two beads: a horizontal C-shape retrace (clean 180° reversal at the top of the loop, `U · U_prev = -1`) and a six-pass zigzag infill. Before the fix the retrace twists visibly at the turn; after the fix the surface is continuous.

## Test plan
- [x] `uv run pytest tests/` — 197 passed
- [x] `npx tsc --noEmit -p jsconfig.json` — clean
- [x] `uv run ruff check . && uv run ruff format --check .`
- [x] Visual check: `uv run python examples/20_toolpath_zigzag.py` shows a clean retrace and infill
- [ ] Visual sanity on an existing tube example (e.g. `11_toolpath.py`) to confirm no regression on the common smooth-curve case

### DIFF
--- a/examples/20_toolpath_zigzag.py
+++ b/examples/20_toolpath_zigzag.py
@@ -1,0 +1,139 @@
+"""
+Zigzag toolpath — twist artifact on hairpin turns (regression test)
+
+A serpentine infill path (parallel lines connected by 180° hairpins) is
+pathological for the current constant-up frame derivation in
+``add_parametric_tube``: when the tangent reverses at a hairpin, the
+width axis ``U = V x T`` also reverses, so adjacent rings point their
+cross-section vertices at opposite physical points. The quad between
+them twists into a figure-eight and you get self-intersecting faces at
+every hairpin.
+
+Two beads are drawn:
+
+1. ``retrace`` — two horizontally-stacked passes (z=0 forward, z=dz back)
+   joined by a short C-shape turnaround that extends past the end point
+   so the tangent stays horizontal through the hairpin. Two intermediate
+   "return loop" points sit side-by-side with opposite tangents, giving
+   a clean 180 deg reversal at a single ring pair.
+2. ``infill`` — a classic zigzag pattern with six parallel passes.
+
+Rotate around the hairpin: with the bug present, the bead pinches and
+flips inside-out where the tangent reverses. After the frame-continuity
+fix, quads stay flat and the tube surface is continuous.
+
+Run: uv run python examples/20_toolpath_zigzag.py
+"""
+
+import time
+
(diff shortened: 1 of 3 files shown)


tdamsma:
Addressed in c978de3.

Copilot's concern was right: the hairpin sweep flips `U` but cap code derived `T = U × V`, so after a flip `T` pointed opposite the real spine tangent and the end cap on a return leg extruded backwards into the tube body. Same failure mode in the static revolution caps, the animated `updateEndCap`, the analytic cap-normal helpers, and the worker mirror.

Fix: carry the already-computed central-difference `tangents[]` array through the build and read `T` from it everywhere caps are constructed, instead of recomputing `U × V`.

- `buildRevolutionCap` (main) + `buildCap` (worker): `T = tangents[spineIdx]`.
- `writeAnalyticCapNormals` / `writeAnalyticCapNormalsW`: take `Tx, Ty, Tz` as args.
- Worker postMessage transfers `tangents`; both `tubeMorphData` creation sites store it.
- `morphFrontierRing` lerps `tx/ty/tz` into `morphedState`; `updateEndCap` and `updateMorphedNormals` read it from there (or from `md.tangents` for un-morphed state).
- `orientations` path (sweep skipped): `tangents[i]` is overwritten with the quaternion's `U × V` so downstream code stays uniform regardless of path.

Checks: `tsc --noEmit` clean, ruff clean, 197 pytest passed.

PR #42, opened 2026-04-22
Miter-joined cross-section frames for parametric tubes

tdamsma:
## Summary

Wide beads around sharp corners used to produce two visible artefacts:

1. A fan of triangles spraying off the **outside** of the corner as adjacent frames rotated through large angles between closely-spaced samples.
2. An inside **crease** where ring inner edges penetrated each other.

Both got worse with denser sampling — densely-sampled corner arcs meant many adjacent frames each rotated through a small angle but sat stacked on top of each other, so the cross-section rings overlapped heavily.

## What changed

Tangent at each interior spine point is now the **unit bisector** of the incoming/outgoing segment directions (not the central-difference average, which diverges for unequal-length neighbours). The cross-section's width axis is **scaled by \`1 / cos(half_turn_angle)\`** so the outer surface stays at constant half-width from each adjacent segment's centerline — classic 2D stroke miter, lifted into the 3D extrusion.

- Past a miter limit of 4 (SVG default), corners fall back to scale = 1 — strictly no worse than the pre-miter build at near-hairpins where the bisector is anyway ill-defined.
- Miter scale is applied only to **position** writes. Normals continue to use the orthonormal \`U, V\` frame, so shading of the miter patch is automatically the bisector of the two adjacent-segment surface normals without any extra work.
- Miter scales are stored alongside local frames so the frontier-morph path during \`draw_range\` animation lerps them between adjacent rings; end caps relocate cleanly to the morphed frontier.
- Single source of truth across main thread and LOD worker: the main-thread \`computeMiterFrames\` is injected into the worker blob via \`Function.prototype.toString()\`, so there's exactly one copy to edit.
- Rolling per-segment direction locals — no \`nSeg*3\` scratch buffer, so the helper is O(1) memory on any spine length.
- \`add_parametric_tube(orientations=...)\` overrides bypass miter entirely — explicit per-point quaternion frames are used as-is and miter scale is pinned to 1.

## Before / after

\`examples/21_tube_corner_zfight.py\` shows the result on a rounded-square path at two regimes:

- **Moderate** (corner radius ≈ bead half-width): miter produces a clean fan of quads, no ring overlap.
- **Aggressive** (radius << half-width): the bead is physically larger than can fit around the arc, so an inside crease remains — that's a real geometric self-intersection no frame trick can fix. The *outside* surface, previously a triangle-fan explosion, is now smooth.

## Performance

Measured on a 1M-point helical spine (\`tmp_bench.mjs\`, pure-JS Node bench of the worker's build path; deleted after use):

| | min | median | max |
|---|---|---|---|
| pre-miter baseline | 302 ms | 304 ms | 327 ms |
| miter | 349 ms | 353 ms | 421 ms |

**+48 ms / +16%** median overhead on a full 1M-point build. All of this lands in the LOD worker (off the main thread) for any tube ≥ 25k points, so it doesn't block the UI at the scale where it matters. Tubes below the LOD threshold build synchronously but are sub-5 ms either way.

## Test plan

- [x] \`uv run pytest\` — 154 passed, 43 skipped
- [x] \`uv run pytest tests/test_browser.py tests/test_parametric_tube.py\` with Chromium — 55 passed
- [x] \`npx tsc --noEmit -p jsconfig.json\` — clean
- [x] \`uv run ruff check .\` — clean
- [x] Launch \`examples/21_tube_corner_zfight.py\` — miter output matches expectation on both regimes
- [ ] Visual regression on \`examples/11_toolpath.py\`, \`examples/15_toolpath_interrupted.py\`, \`examples/18_parametric_tube.py\`


tdamsma:
Self-review pass — addressed the follow-ups in `ecc6213`:

- **Single source of truth for `computeMiterFrames`.** The worker copy is gone; the worker blob now injects the main-thread function's source via `${computeMiterFrames.toString()}`. `TUBE_MITER_LIMIT` is already defined in both scopes, which is the only symbol the injected source references. Browser tests (55 passed on Chromium) confirm the Worker receives and runs the injected source correctly.
- **No more `nSeg*3` scratch buffer.** Rewrote the helper with rolling `prev`/`curr` segment-direction locals, so it's O(1) memory on any spine length — drops ~12 MB throwaway on the 1M-point bench path.
- **PR body cleanup.** Removed the checklist item referencing the non-existent `examples/20_toolpath_zigzag.py`; added the browser-tests-pass line.

Net: +65 / -157 across `viewer.js` + `viewer.html`. All prior checks still pass (pytest 154p/43s, tsc clean, ruff clean).

Deferred (not in scope for this PR, flagged for follow-up):
- V-axis miter correction for turns in the vertical plane — current code only scales U, which is exact for planar-horizontal toolpaths but degrades silently for vertical zigzags.
- Dedicated Playwright assertions on miter ring positions — existing tube tests exercise the path end-to-end but don't pin the geometry.

tdamsma:
Merged main into the branch — the conflict was with PR #41's hairpin-reversal fix, which landed after we branched.

**How the two fixes compose:**
- `computeMiterFrames` now also outputs a tangents array (the unit-bisector tangent it already computes). That replaces PR #41's central-difference pass — one source of truth for tangents.
- Orientations path fills `tangents[i] = U × V` from the quaternion, matching PR #41's behaviour.
- Hairpin U-flip sweep runs after `computeMiterFrames`, only when `!orientations`. When rewriting a flipped ring, it re-applies `miterScales[i]` so mitered corners survive the flip.
- `tangents` + `miterScales` both carry through the LOD worker transfer list, `tubeMorphData`, and the `morphedState` lerp. U-flip is invariant in the tangent; miter scale is invariant in U — so the two stay independent.

`d3a5d19` on the branch. `uv run pytest` 197/0, tsc clean, ruff clean. The PR is mergeable again.

PR #43, opened 2026-04-26
Collapse self-intersecting strand folds in parametric tubes

tdamsma:
## Summary

Adds `strand_collapse=True` flag on `add_parametric_tube`. When the local curvature κ exceeds 1/half-width, the inner offset surface of the bead self-intersects geometrically and the renderer paints overlapping triangles. The flag runs a per-cross-section-vertex fold detector + snap pass on the LOD worker, returning crisp creases instead of self-intersecting fans, with the bead's top-view footprint preserved by construction.

Off by default, opt-in per call. Non-blocking — the main thread never waits on the O(nCs · nSpine · WIN) scan.

## Algorithm

Per cross-section vertex k (independent strand) — six per ring on the chamfered hex:

1. **Build the un-mitered strand polyline:** `strand[i] = spine[i] + u·U[i] + v·V[i]` from the local `(u, v)` for vertex k. Detection runs on un-mitered geometry — miter-scaled U at sharp corners can synthesise spurious local minima that aren't real folds.
2. **Compute 3D seg-seg shortest-line distance²** for every pair `(i, j)` with `j − i ∈ [4, 50]`.
3. **Fold target = strict local minimum** of that distance² grid (vs. its 8 in-window neighbours) AND below `(0.04 · max(W, H))²`.
4. **Snap rings `(i+1..j)`** of strand k to the closest-pair midpoint of the *mitered* mesh segments — the un-mitered strand defines the geometric truth, but the snap target lands inside the real mesh.

**Coverage preservation** is structural: only inside-of-bend strands satisfy the local-min test (the outside strand's seg-seg distance is monotonically large). The outside tip strand defines the top-view silhouette and is never moved, so the bead footprint can't shrink.

**Snap-only, no re-triangulation.** Coincident strand vertices give zero-area triangles in the two columns adjacent to k; the GPU culls them in hardware. Index buffer, normals (cross-section bisector), and colors stay intact.

## Architecture

- `buildParametricTubeGeometry` no longer takes a `strandCollapse` arg. Tube creation is purely synchronous geometry construction.
- After mesh creation, when `strand_collapse=True`, the main thread posts a `'collapseOnly'` message to the existing LOD worker with cloned positions + spine + widths + heights + localFrames. The worker runs `collapseTubeStrandFolds` and transfers the snapped positions back; the main thread copies them into the BufferAttribute and marks `needsUpdate`.
- For LOD-enabled tubes (≥ 25k samples), the same `collapseTubeStrandFolds` is inlined into `_LOD_WORKER_CODE` and runs on every reduced-spine rebuild, so creases stay crisp at every camera distance.
- A spine-distance pre-filter (`kMax`) skips segment pairs whose endpoints sit further apart than the worst-case fold radius — saves ~10–12% on realistic-density tubes.

## How to use

```python
v.add_parametric_tube(
    "bead",
    spine=spine,        # (N, 3) float32
    widths=widths,      # (N,) float32
    heights=heights,    # (N,) float32
    strand_collapse=True,
)
```

Composes with all other `add_parametric_tube` kwargs (`colors=`, `lod=`, `anchor=`, `orientations=`, …) and with `add_toolpath(**kwargs)`.

UX note: the flag is async. Tube creation returns immediately with the un-mitered (un-collapsed) mesh; collapsed positions arrive when the worker finishes. Round-trip is < 10 ms for ≤ 1k tubes, ~150 ms at 10k, ~1.3 s at 130k — none of it on the main thread.

## Bench

`bench/strand_collapse_bench.mjs` (V0…V3, realistic 10 samples/segment, loop scaled with `nCtrl` so density stays constant):

```
N         | nCtrl  | V0 ms     | V1 (kMax)       | V2 (hoisted)    | V3 (streaming)
----------+--------+-----------+-----------------+-----------------+-----------------
      130 |     13 |    1.10   |   0.762   1.46× |   0.391   2.83× |   0.399   2.78×
     1300 |    130 |   14.39   |   12.82   1.12× |   12.82   1.12× |   12.93   1.11×
    13000 |   1300 |  143      |  128      1.11× |  129      1.11× |  130      1.10×
   130000 |  13000 | 1426      | 1289      1.11× | 1278      1.12× | 1286      1.11×
```

Production ships V2 (V1 with the kMax check hoisted out of the inner loop). V3's 3-row sliding dist² buffer doesn't help — V8 + CPU caches handle the 24 MB sequential grid fine; residual cost at large N is the polylineSegSegMidpoint compute itself.

The big UX win is architectural: every `*` ms above is now off the main thread. Tube creation that previously froze the renderer for 1.3 s now returns in ~1 ms (postMessage overhead).

## Demo

`examples/21_tube_corner_zfight.py` — three-tube oracle:

- LEFT: baseline `add_parametric_tube` with sphere markers at every fold target detected by the Python prototype.
- MIDDLE: `add_parametric_tube(strand_collapse=True)`. Worker runs the snap pass.
- RIGHT: chamfered-hex tube built locally in Python with the same snaps applied (visual oracle for the JS implementation).

The middle and right should match at every collapsed corner. Open serpentine spine with rising Z and sinusoidal width, exercising 17 folds across all 6 cross-section vertices.

## Test plan
- [x] `uv run pytest` — 199/199
- [x] `npx tsc --noEmit -p jsconfig.json` clean
- [x] `uv run ruff check . && uv run ruff format --check .` clean
- [x] `uv run python src/threejs_viewer/viewer/build.py` rebuilds `viewer.html`
- [x] `node bench/strand_collapse_bench.mjs` — V1/V2/V3 outputs match V0 bitwise
- [ ] Manual: `uv run python examples/21_tube_corner_zfight.py` — middle and right tubes show identical creases at all four corners; baseline (left) shows visible z-fighting fans.


tdamsma:
Addressed the Copilot review in 709e5eb:

**Fixed in this push:**
- **`collapseOnlyResult` race** — load token now snapshotted at send, round-tripped through the worker, verified on receive via `_isLoadTokenCurrent`. Late results from a tube that's been deleted (or deleted-then-re-created with the same id) drop out instead of stomping the new mesh.
- **`init`/`dispose` ordering** (suppressed-low-confidence, but the bug was real) — `_deleteObject` now runs before the worker `postMessage` calls, so the worker queue is dispose(old) → init(new) → collapseOnly(new). Previously the trailing dispose for the old tubeLOD could clobber the new tube's worker state because both shared the same `tubeId`.
- **Unused `mesh.userData.tubeStrandCollapse`** — removed.

**Stale (no fix needed):**
- *Bidirectional scan / point-to-segment asymmetry on `viewer.js:961`* (commit `db2b281`) — the symmetric-scan fix shipped in `ca9e68b`, and the algorithm rewrite in `ec84b0e` replaced the entire point-to-segment detector with a 3D seg-seg shortest-line local-min test that has no asymmetry to begin with. The line range cited no longer exists in current code.
- *`catmull_rom_closed` naming on `examples/21_tube_corner_zfight.py:44`* (commit `47a108f`) — renamed to `catmull_rom_periodic` in `ca9e68b`. The current example also defines `catmull_rom_open` and uses it for the open serpentine path.
- *Missing `strandCollapse` header tests on `client.py:957`* (commit `1bb1c34`) — already covered by `tests/test_parametric_tube.py:191-204` (`test_add_parametric_tube_strand_collapse_default_omits_header_key` and `test_add_parametric_tube_strand_collapse_true_sets_header`).

199/199 tests pass, type-check clean.

PR #54, opened 2026-06-02
docs: changelog for distance-fog dynamic-range fitting

tdamsma:
## What

Documents the **distance-fog dynamic-range fitting** that shipped in `0.0.32` but went undocumented.

The fog near/far fit (`DepthCueController._updateFogRange()`) was developed as a follow-up to the `0.0.31` line-depth-cue feature, but its working-tree edits got folded into the picking commit (`cdcb80d`, #53) when that PR was committed — so the code is already on `main`, just unrecorded in the changelog. Worse, the `0.0.31` entry still describes the **old** scene-sphere-proxy behaviour (`near/far track the scene bounding sphere and camera distance`) that this change replaced.

This PR is **changelog-only** — no code changes (the code is already on `main`).

## The shipped behaviour being documented

- Fog near/far are now fitted each frame to the **closest→furthest view-space depth of the visible polylines** (per-polyline bounding-box corners in `-mvPosition.z`), instead of the whole-scene bounding sphere.
- `near` is anchored to the closest visible line; `far` is pushed *past* the furthest by `FOG_DEPTH_COVERAGE` (0.4) so the content maps into the **toe** of the smoothstep ramp — the bulk of the lines stay near the light end (~70% under ~0.2 fog) with darkening concentrated on the farthest stretch (furthest ≈0.35), rather than the old behaviour where a far-away or depth-thin drawing washed out to a near-uniform dim.
- Falls back to the scene-sphere proxy when no line geometry is visible.

## Notes

- Left the `0.0.31` entry as the historical record of what `0.0.31` shipped; the change is documented under `0.0.32` where it actually landed.
- A separate follow-up PR will scope the depth cues (fog + EDL) to polyline geometry only (per `plans/depth-cue-scope-to-toolpath.md`).


tdamsma:
Folding this into the single PR for the depth-cue work (scope cues to polylines). The fog dynamic-range CHANGELOG entry is now included there, so this changelog-only PR is redundant.

PR #60, opened 2026-06-22
Move gizmo: public attachMoveGizmo + per-axis setGizmoAxes for embedders

tdamsma:
Closes the threejs-viewer-side API asks for the ribweaver `/cells` gizmo migration (`ribweaver/docs/threejs-viewer-gizmo-asks.md`), letting embedders drive the viewer's shared `TransformGizmoController` instead of hand-rolling `TransformControls`.

### 1. `viewer.attachMoveGizmo(object3D, id=null)` (Required)
Public surface to attach the move gizmo to **any** `Object3D` — including bare sentinel handles that were never registered in `_objects` (which `enableMoveGizmo({id})` can't reach). It ensures-enabled via a shared `_activate()`, so a programmatic attach gets the modifier keys, the per-frame `update()` loop (camera tracking, stale-object auto-detach), and move reporting live — exactly like an interactive attach. JS-only: an embedder driving bare `Object3D`s does so from the browser, and Python has no handle to them.

### 2. `setGizmoAxes` / `set_gizmo_axes` (Required)
Constrain which translate/rotate axes the gizmo exposes (drives `TransformControls.showX/Y/Z`) — e.g. `{x:false, y:false, z:true}` for a Z-only rail or placement floor-mode masking. Wired through all four layers:
- JS controller `setAxes(mask)`
- public `viewer.setGizmoAxes({x,y,z}|null)`
- `set_gizmo_axes` WebSocket message
- Python `set_gizmo_axes(x=True, y=True, z=True)`

An axis shows unless its key is explicitly `false` (so the JS surface matches the `x=True` Python defaults; every fully-specified call behaves identically). `detach()` resets to all-axes so a constraint can't silently leak to the next target; the Python state replays on reconnect **after** the enable, and clears on `disable_move_gizmo`.

### 3. Alt-held mode is now a true momentary override
Holding Alt switches the live control to rotate; releasing it falls back to the **caller-set base** `this.mode` instead of hard-resetting to `'translate'`. The persistent base stays distinct from the transient mode, so an Alt tap can't clobber a `setGizmoMode('rotate')` — yet the default translate↔rotate toggle still works. (This is the *correct* form of the doc's suggested one-liner, which would otherwise stick the gizmo in rotate after one Alt tap from a translate base.)

## Test hardening (pre-existing flakiness)
The gizmo browser tests flaked ~25% when run together (reproduced on the **original, untouched** tests) — they asserted immediately after a fixed `time.sleep` on async WS round-trips. Replaced with deterministic condition-polling (`_wait_for` via Playwright `wait_for_function`, `_wait_until` for receive-thread state), cutting the run-together flake rate to ~4% over 26 stress runs (the lone residual was an unreproducible synthetic mouse-drag/click miss through `TransformControls`' raycaster, unrelated to these API changes).

## Verification
- Unit tests: **188 passed** (new: `set_gizmo_axes` payload/defaults + disable-clears-constraint)
- Full browser suite: **62 passed** (new: untracked-`Object3D` attach, Alt momentary-override, `setGizmoAxes` constrain+reset-on-detach)
- `ruff check` + `ruff format --check`: clean · `tsc --noEmit -p jsconfig.json`: clean · `viewer.html` regenerated via `build.py`


tdamsma (commented):
Approving with nits. This cleanly lands all four asks from `threejs-viewer-gizmo-asks.md`:

- **`attachMoveGizmo`** — the required bare-`Object3D` attach, with `_activate()` auto-enabling so reporting / modifiers / `update` all go live; verified against an untracked sentinel.
- **`setGizmoAxes`** — across all four layers (JS controller → public method → WS case → Python client) with proper detach-reset + reconnect-replay.
- **Alt-momentary fix** — in the *correct* form: writes the control directly and keeps `this.mode` as the base, rather than the mutate-`this.mode` one-liner the doc suggested.
- The `sleep`→poll test hardening is a welcome bonus and honestly scoped.

Two trivial Copilot nits worth folding in:
1. Wrap the `_wait_for` predicate in try/catch so a pre-construction throw counts as "not ready" rather than failing the wait.
2. Either return the predicate value from `_wait_until` or fix its docstring (it says "returns the result" but returns `True`).

For the downstream cells gizmo migration: the **TCP gizmo** and the **placement floor-mode axis mask** are now fully unblockable. The placement gizmo's *relative-to-drag-start* 100 mm snap and per-frame live-numeric decompose still want a thin hand-roll (or a future relative-snap / pre-report hook), since `onObjectMove` is report-only — but that's out of scope here and not in the asks doc.

Ship it.

PR #80, opened 2026-07-03
Point clouds: per-point time windows + streamed octree LOD (#79)

### LINKED ISSUE #79: add_points: distance-dependent LOD (octree) + per-point time-threshold visibility for large animated voxel fields
## Context

mill-sim wants to push a **dense voxel field** (the machining stock band around a
part) to the viewer as a `THREE.Points` cloud and **animate material removal**
over time with a scrubber. Target sizes: **1M points now, up to ~5M**. Each voxel
carries the timestamp at which the cutter removes it (voxels never cut have no
timestamp and stay forever).

Two gaps block this today.

### 1. Distance-dependent LOD for large point clouds (octree / Potree-style)

`add_points` renders the whole cloud in one draw call, which is great for a
static 1M cloud, but at 5M — and especially with `size_attenuation` — near
geometry oversaturates and far geometry wastes fill rate. Potree solves this with
an **octree that streams/render coarser density at distance**.

**Ask:** an opt-in LOD mode for `add_points` (build an octree over the positions,
render a camera-distance-adaptive subset). Should preserve per-point scalar
colours and the animation channels below. Open design Qs: build octree on the
Python side vs in-browser; memory budget; whether we adopt the Potree format or a
lighter in-house octree.

### 2. Per-point time-threshold visibility (out-of-order removal)

Material removal is **not** a monotonic prefix of the buffer, so the existing
`set_draw_range` reveal (leading `frac*N` points) can't express it. What we need:

- a per-point scalar attribute (e.g. `removal_time`, NaN/∞ = never removed), and
- a global `time` uniform where the shader **discards points with
  `removal_time <= time`** (i.e. already cut).

Driven by a scrubber this gives a true removal animation without pre-sorting the
buffer. `clip_times` today only seeks an embedded GLTF AnimationMixer clip — it is
not per-point point-cloud discard — so this is a new channel.

_(These two are separable — happy to split #2 into its own issue if preferred.
Filing together because both are needed for the same "animated voxel field" use
case in mill-sim's interactive spec.)_

🤖 Filed from the mill-sim interactive-spec session with [Claude Code](https://claude.com/claude-code)

tdamsma:
Implements both halves of #79 for the in-RAM tier, plus the decision doc that settles the architecture (`plans/points-octree-lod.md`, synthesized from five prior-art surveys: Potree ecosystem, formats/Python tooling, LOD techniques, encoding/compression, GPU efficiency).

## 1. Per-point time-window visibility

`add_points(birth_times=, removal_times=)` — optional (N,) lifetimes; the vertex shader shows a point only while `birth_time <= t < removal_time`, culling by pushing `gl_Position` outside the clip volume (pre-raster, zero fragment cost — Potree's GPS-time-filter trick via `onBeforeCompile` + `customProgramCacheKey`). Expresses the **out-of-order** appearance/removal that `draw_range`'s buffer-prefix reveal cannot. NaN/±inf ends = unbounded (mapped to ±FLT_MAX Python-side; GLSL NaN compares are undefined).

Scrub time drivers: `set_points_time(id, t)` (streaming) and the new **`point_times` animation channel** (`Animation.set_point_time_data`, `Frame(point_times=...)`) — dragging the animation slider scrubs points in/out.

| t=0 | t=5 | t=9.5 |
|---|---|---|
| t0 | t5 | t9.5 |

*A 1-unit-wide birth/removal band sweeps a 40×20 grid; the NaN-unbounded bottom rows persist. At t=9.5 every windowed point is gone.*

## 2. Streamed octree LOD

`add_points(lod=True | {node_capacity, point_budget, refine_pixels, max_depth, seed})`:

- **Build (Python, `points_lod.py`)**: Potree-style *additive sampled octree* — every node owns a seeded pseudo-random sample of its cube, **time-stratified** over birth/removal so the time filter thins every LOD level uniformly (sampled real points, never aggregated rollups: any merged voxel would fabricate a lifetime belonging to no real point). Per-node vertex order Morton-sorted + 128-batch shuffled (Schütz 2021, up to ~4× raster throughput).
- **Transport**: one WS header + a 40 B/node binary hierarchy (bounds, count, `[min birth, max removal)`, BFS topology). Node payloads are **lazy callable blob-store providers** — each `GET /points_lod_<uuid>/<i>` quantizes + packs a numpy slice on demand; nothing materialized up front. Payload: normalized **int16 xyz** over the node cube (free GPU dequant via normalized attribute + node transform; ints stay GPU-resident, unlike Potree's CPU float decode) + u8 rgb + f32 times.
- **Viewer runtime**: per-frame priority traversal (max-heap on projected node size) — refine while ≥ `refine_pixels`, hard stop at `point_budget` (default 1.5M), ≤6 concurrent node fetches, LRU eviction past ~3× budget, per-node frustum culling, and **per-node time culling** (a node whose window misses the scrub time is neither drawn nor fetched). The group shares one scrub-time uniform, so `set_points_time`/`point_times` drive LOD clouds identically to flat ones. `draw_range` is a documented no-op on LOD clouds (buffer order is per-node Morton).

| far view, t=8 (656k pts, level ≤3) | zoomed, t=8 (1.49M ≤ 1.5M budget, level 4 streamed in) | far view, t=60 (only NaN bedrock left) |
|---|---|---|
| far | near | t60 |

*2M-point wavy sheet: growth front + eroded trailing edge at t=8; zooming streams 54→102 nodes under the budget; at t=60 nodes time-cull out (95 visible of 114 loaded) and the shader keeps only bedrock.*

## Bugs caught by verification before landing

- **Child octree cubes never shrank** (`quarter * 2.0` for the child half) — every node got the root-sized cube, inflating projected sizes (everything refined) and exploding tree depth to `max_depth`. Regression test asserts `half == root_half / 2^level`.
- 8× coarse-node point-size boost produced giant occluding sprites near the camera (coarse ancestors render close-up under additive refinement) — capped at 2×, √2 per level.

## Testing

- 342 tests pass: 15 new builder/protocol tests (permutation/tiling/cube-halving/topology/time-stratification/quantization round-trip/blob lifecycle), 10 new Phase-1 protocol/animation tests, 3 new Playwright tests (shader compile watchdog, `point_times` channel from the playhead, LOD streaming within budget + time-cull-out).
- End-to-end headless-Chromium verification scripts with the screenshots above.

## Out of scope (Phase 3, blueprinted in the plan doc)

Public `node_provider` hook for the ~1B provider-synthesized tier (the callable-blob mechanism already supports it), u16 time quantization (needs an immortal-flag sentinel), optional blue-noise sampling, Morton-radix builder speedup (current: ~1M pts/s, fine to ~20M).

Closes #79.


tdamsma:
All three Copilot findings addressed in 307076b:

1. **Budget `break` under-fill** — the traversal now `continue`s past a node that exceeds the remaining budget (skipping its subtree, since children without the parent's coarse sample leave density holes) and keeps popping, so smaller branches elsewhere still fill the budget.
2. **Payload bounds** — `add_points_binary` validates the full expected byte count (positions + optional colors/birth/removal blocks) before parsing; a truncated payload now fails loudly instead of silently zero-padding the time blocks.
3. **`lod={}` falsey** — explicit `is not None and is not False` check; an empty options dict enables LOD with defaults (regression-tested).

Stacked branches (#81, #82) rebased on top; full suite green (364 tests incl. browser).

PR #94, opened 2026-07-06
Travel-move rendering: add_toolpath(travel="line") + line-segments polyline (#88)

### LINKED ISSUE #88: Travel-move rendering for toolpaths: line-segments primitive or add_toolpath(travel=...) option
## Context

`Toolpath.from_gcode` + `add_toolpath` handle travel moves well on the bead side: zero-width points split the path into separate extrusion tubes with tapered caps (documented in `examples/15_toolpath_interrupted.py`). But there is no supported way to *show* the travel moves themselves — the thin "travel lines" every slicer preview draws.

## What's missing

A real G-code toolpath has many disjoint travel episodes (a large-format file we drive has 2,222 of them, one per layer seam). Options today:

1. **One `add_polyline` per episode** — thousands of objects/draw calls, and no practical way to animate them in sync with the bead's `draw_ranges` channel.
2. **One polyline over the whole path** (what we ended up doing in nc-gcode-interpreter's `nc-view`): a single native line driven by the *same* `draw_ranges` fractions as the tube. On extrusion stretches the line hides *inside the opaque bead*, so only travel hops show, perfectly in lockstep.

Workaround 2 works nicely but relies on occlusion as a side effect rather than a contract:

- it breaks as soon as the tube gets `opacity < 1`;
- it can peek through where LOD thins the tube below the spine;
- nothing in the docs suggests it, so every consumer has to reinvent it.

## Proposals (either would do)

- **`add_toolpath(..., travel="line")`** (or `travel_color=`): the group builds the travel line itself from the zero-width stretches it already computes, and wires it into the group's draw-range distribution. Highest-level, matches how `from_gcode` already owns travel detection.
- **A native line-*segments* primitive** (`add_polyline(..., segments=True)` → `THREE.LineSegments`): lets a consumer draw exactly the travel edges as one object without false connectors between episodes. Needs a story for `draw_ranges` (reveal by segment index; a `bias_index`-style mapping to original spine indices would keep it in sync with the tube's fraction-of-points semantics).

## Also worth documenting

That a single `draw_ranges` channel can drive multiple objects of different types in lockstep (`set_draw_range_data([tube_id, line_id], data)`) — it works today (`makeChannelApply` dispatches per `userData` type) and is exactly what makes the workaround viable, but no example shows it.

Happy to contribute either variant — the whole-path-line workaround lives in `nc-gcode-interpreter/python/nc_gcode_interpreter/viz.py` (`view_toolpath(travels=True)`) if you want the reference implementation.

tdamsma:
Closes #88.

A real G-code toolpath has many disjoint travel episodes; the only supported ways to show them were one object per episode (thousands of draw calls, unsyncable) or a whole-path line hidden inside the opaque bead — occlusion as a side effect, broken by `opacity < 1` and LOD thinning. This lands **both proposals** from the issue, composed:

## `add_polyline(..., segments=True)`
Disjoint line *segments* (`THREE.LineSegments`): consecutive point pairs each draw an independent edge — any number of separate episodes in one object/draw call with no false connectors. Even point count enforced Python-side; implies the native path (`fat` ignored) and is never pickable (an edge soup has no arc length). Per-vertex colors and `set_draw_range` (leading `frac*N` points ⇒ whole edges) work as usual.

## `add_toolpath(..., travel="line", travel_color=...)`
The group builds the travel line itself from the zero-width stretches it already computes: every spine edge **not interior to an extrusion run**, so each hop spans from the last extruded point through the zero-width points to the next extruded point (the line meets the bead ends). One `<id>_travel` segments child, wired into the group's draw-range distribution **by contract**: `register_toolpath_group` carries `travelId` + ascending `travelEndFracs` (global spine fraction of each edge's end point, same `/n_total` convention as `segmentRanges`), and `applyToolpathGroupDrawRange` binary-searches them to reveal whole hop edges in exact lockstep with the beads. A single-run toolpath with travel now takes the group form when the line is requested (the bare-tube fast path has nowhere to hang the mapping); without travel stretches the option is a no-op.

Also documented (per the issue): one `draw_ranges` channel can drive **multiple objects of different types in lockstep** (`set_draw_range_data([tube_id, line_id, cloud_id], data)`).

`examples/15_toolpath_interrupted.py` now shows the travel line riding the existing draw-range animation.

## Tests
Protocol: segments header/constraints, travel-edge fractions (ascending, `/n_total`), single-segment-with-travel grouping, validation + no-travel no-op. Browser: `LineSegments` child mounts under the group and the reveal count steps exactly at each edge's end fraction (0 → 2 → 6 → 10 verts across the sweep).

Screenshots incoming in a comment (62% reveal with partial hops, 100% with all four layers' hops).


tdamsma:
Visual check (example 15's pill toolpath, `travel="line", travel_color=0xFF8800`):

**62% reveal** — earlier layers' bottom-straight hops are in, the current layer's not yet (whole edges appear only once the global spine fraction passes each edge's end point, in lockstep with the bead reveal):

travel line at 62%

**100% reveal** — all four layers' hops, one `LineSegments` draw call:

travel line at 100%

(Screenshots on the [pr-94-images](https://github.com/CEAD-group/threejs-viewer/releases/tag/pr-94-images) pre-release.)

PR #98, opened 2026-07-06
Playback pacing + frontier sync: EMA-paced playhead, offset uniform-dt, group range convention, LOD remap (#96, #97)

### LINKED ISSUE #96: _getFrameAtTime uniform-dt fast path assumes frame times start at 0
`_getFrameAtTime` has a fast path for uniformly spaced frames:

```js
if (this._animation.uniformDt > 0) {
    const raw = time / this._animation.uniformDt;
```

`raw` is only a frame index if `frames[0].time == 0`. For a uniformly spaced animation whose times start at some offset (e.g. `set_frame_times(100000 + arange(n) * 0.01)`), every lookup lands at `raw >> lastIdx`, so the playhead clamps to the last frame: seeking anywhere shows the end state.

Repro (playwright, against a stock viewer):

```python
times = 100_000.0 + np.arange(200) * 0.01   # uniform, offset start
anim = Animation(loop=False)
anim.set_frame_times(times)
# transforms: x slides 0 -> 1 across frames
v.load_animation(anim, autoplay=False)
# seek to the middle:
page.evaluate("() => window.threejsViewer._seekToTime(100001.005)")
# object x == 1.0 (last frame) instead of ~0.5
```

Fix is one term: `const raw = (time - frames[0].time) / uniformDt;` (and the same offset in any other consumer of `uniformDt`).

The non-uniform binary-search path is unaffected — I verified it stays float64-exact end to end at 497k frames spanning 1e5 s (storage, seek, and transform interpolation all within 0.01 point-widths of numpy float64 expectations), which might be worth keeping as a regression test; happy to PR both the fix and the tests.

### LINKED ISSUE #97: Playback jerk on heavy scenes: playhead advances by raw wall-clock delta, so frame-cost jitter becomes velocity jitter
## Symptom

Playing a large toolpath animation (497k frames, ~2,200 tube segments, real 46 MB NC program) at a high speed multiplier, playback is visibly jerky — sudden accelerations/decelerations of BOTH the bead frontier and a transform-driven nozzle marker, in lockstep. Gets worse as the print grows. Independent of camera tracking (T off: no change) and zoom level.

## What it is NOT (measured, headless chromium)

- Not time precision: frame times survive to the browser as float64 end to end. At 497,818 frames spanning 1e5 s with heterogeneous dts, seeks from 5%–97% of the timeline reproduce numpy-float64-exact interpolated positions to within 0.01 point-widths.
- Not the frame lookup or channel interpolation: during playback the applied transform tracks `_animationTime` to 0.0015 anim-seconds (constant-velocity test object, 150 rAF samples at 200x).

## What it is

`_animate()` advances the playhead by raw wall-clock delta:

```js
const deltaTime = (now - this._lastAnimationUpdate) / 1000;
this._animationTime += deltaTime * this._animationSpeed;
```

Sampling per-tick playhead advance in a MINIMAL headless scene (one box, 497k frames, 200x):

```
per-tick anim advance: median 2.14 s, p95 3.05 s, max 9.38 s
```

One slow render frame = a 4x playhead jump. On a real scene the frame cost fluctuates continuously (tube frontier morph, LOD, growing geometry), so playback velocity fluctuates proportionally — perceived as jerk with sudden accel/decel, amplified linearly by the speed multiplier. The animation position is always CORRECT for the playhead time (mapping is exact); it's the playhead's advance rate that is noisy.

## Proposal

Keep wall-clock as the long-term reference but stop passing per-frame noise through:

- Clamp the per-tick delta (e.g. `min(deltaTime, 3 * smoothedDelta)`) so a single stalled frame cannot teleport the playhead, and/or
- Smooth the delta with a short EMA (~10 frames) before applying the speed multiplier — playback velocity then stays visually constant while remaining tied to real time on average.

Optionally expose a `playback_pacing: "smooth" | "realtime"` knob if exact wall-clock coupling matters to some consumer.

## Repro

```python
# 497k frames, mixed dts (31 ms / 125 ms / 250 ms / 2 s), one box moving at
# constant true velocity (x proportional to time); seek to 30%, play at 200x,
# sample (v._animationTime, box.x) every rAF for 150 frames.
# -> x tracks _animationTime to 0.0015 anim-s (exact), but
#    diff(_animationTime) has max/median ratio > 4 even headless.
```

Happy to PR the delta smoothing.

tdamsma:
Closes #96, closes #97 — including the real desync behind the #97 report's screenshot, which turned out to be two additional position bugs, both found by measuring the actual 44 MB NC program headless.

## The measured story (havoc_full.mpf, 497,818 points, 200,009 s)
Sweeping the reported playhead (192,401.89 s ± 40 s) and comparing the frontier-ring centroid against the float64-exact toolpath tip:

| | worst frontier-vs-true error |
|---|---|
| before | **531 mm** (segment had `lod=False` — not an LOD artifact) |
| after | **3.8 mm** (float32 channel quantization, sub-bead-width) |

## Fix 1 — group draw-range convention (the dominant desync)
Producers send `draw_range` values as `value = fractional_point_index / (n−1)`; `register_toolpath_group` built `segmentRanges` with `/n`. Recovering the frontier's local index gave `value·n − s` instead of `value·(n−1) − s` — a systematic frontier lead of exactly `value` **points**. One point = one G-code segment: invisible in dense clusters, hundreds of mm on long raster moves near the end of a path. `segmentRanges` + `travelEndFracs` now use `/(n−1)`. Only travel-split group toolpaths were affected.

## Fix 2 — tube-LOD frontier remap under RDP collapse
`remapDrawRangeToReducedPairs` interpolated across a kept span **by index**; RDP collapses collinear runs regardless of spacing, so one long segment + a dense flatten-tolerance cluster put the frontier almost anywhere on the span (synthetic proof: **98 mm** error on a 106 mm run, frontier at x=1.74 for a point at x=100.0). The remap now takes the true position from the retained `originalSpine` at the fractional index and projects it onto the reduced chord — error bounded by the RDP epsilon the geometry is already drawn at.

## Fix 3 — playback pace smoothing (#97's original proposal)
The playhead advances by a ~10-frame EMA of the wall-clock delta instead of the raw delta; single out-of-band deltas are clamped to 3× the running average (a stalled frame can't teleport the playhead), and 3 consecutive same-side out-of-band deltas snap the EMA to the new regime (background-tab throttling recovers immediately in both directions — Copilot's review finding and its mirror image, which CI caught).

## Fix 4 — offset uniform-dt fast path (#96)
`(time − frames[0].time) / uniformDt` — a uniformly spaced timeline starting at t=100000 previously clamped every seek to the last frame.

## Tests
Regressions for all four: LOD frontier lands at x≈100 (not 1.74) on a collapsed collinear run; group frontier lands on the exact addressed point early and late in a mixed-spacing travel-split path; offset-start seek with the fast path active; EMA live during playback + faked 5 s stall clamped; travel-line lockstep thresholds updated to the `/(n−1)` convention.


tdamsma:
Re the two Copilot comments on `_smoothedFrameDelta` never resetting: addressed in the same push by the regime-change snap — 3 consecutive same-side out-of-band deltas re-seed the EMA to the live delta (both directions: returning from a throttled background tab snaps DOWN within 3 frames instead of overrunning the clock while the EMA decays, and an unrepresentatively small seed snaps UP — the latter was caught by CI as a real failure of the initial clamp-only version). Single outliers stay clamped, so a lone stalled frame still can't teleport the playhead. Covered by `test_playback_pacing_smooths_and_clamps_stalls`.

PR #106, opened 2026-07-08
Exact reference colormaps: 256-entry viridis/plasma/turbo tables

tdamsma:
## Problem
Python-side scalar colouring (`ViewerClient._apply_colormap`) approximated viridis/plasma/turbo with ~8–11 hand-picked stops, linearly interpolated. viridis survives that (it's near-linear), but **plasma and turbo don't** — the chord between sparse stops cuts the corners off their non-linear features.

Measured against the true 256-entry reference tables (1000 samples along each ramp, error in 8-bit RGB levels; ~2–3 levels ≈ just-noticeable):

| colormap | mean ΔRGB | max ΔRGB | % of ramp >5 levels off |
|---|---|---|---|
| viridis | 2.3 | 13.2 | 7% |
| plasma | 25.5 | 54.2 | 98% |
| turbo | 30.7 | 68.0 | 99% |

So plasma/turbo were a visibly wrong colour across nearly the whole range.

## Fix
- **New `_colormap_data.py`** — the exact 256-entry matplotlib reference tables (viridis/plasma CC0, turbo Apache-2.0), as `float 0..1` RGB grids. Docstring records the one-line regeneration recipe. The hand-laid 3-per-row grid is wrapped in `# fmt: off` so ruff doesn't explode it to one float per line.
- **`client.py`** — `_apply_colormap` now reads those tables and interpolates between 256 entries instead of the inline ~10-stop lists (net *fewer* lines in client.py). Same interpolation code, same runtime cost, no API change — it's free to be exact.
- **`tests/test_colormap.py`** — pins first/mid/last entries against the true reference values (e.g. viridis node 128 = the real teal `#21918c` the approximation never hit).

## Notes
- No viewer/JS changes — this is purely the Python colour path.
- CHANGELOG deferred to the next release's changelog PR, per repo convention.


tdamsma:
Addressed the Copilot review comments in 5fc4283:

- **Coarse toolpath colormaps** — `toolpath.py` now colours beads through the exact 256-entry `_colormap_data.TABLES` (removed the 11-stop `_COLORMAPS` approximation), so bead colouring matches `ViewerClient._apply_colormap`.
- **float64 tables** — `TABLES` are now float32 and `client._apply_colormap` keeps the interpolation single-precision, dropping the redundant float64→float32 pass over large color arrays.
- **turbo not pinned in tests** — `test_colormap_tables_are_exact_reference_data` now asserts turbo first/mid/last entries.
- **Apache-2.0 notice** — added `THIRD_PARTY_NOTICES.md` (turbo © Google 2019, Apache-2.0, full license text; viridis/plasma CC0), shipped in both the sdist and the wheel (verified the wheel contains `threejs_viewer/THIRD_PARTY_NOTICES.md`), and referenced from the `_colormap_data` docstring.

PR #109, opened 2026-07-08
Lighting panel: environment-map on/off toggle (uglier-but-faster perf knob)

tdamsma:
## What

Adds a checkbox to the **Lighting panel** (press `E`) that toggles the IBL environment map. Off nulls `scene.environment` — dropping the per-pixel PBR reflection lookups — for a **flatter but faster** render; on restores the retained PMREM map without a rebuild. The env-intensity slider disables while the map is off.

Underlying wish: a mechanism to trade some prettiness for perf.

## How it fits the existing lighting plumbing

- **Retained map**: `_loadCubemap` stashes the PMREM texture on `this._envMap` and only attaches it to the scene when enabled, so the toggle is a pointer swap, not a rebuild.
- **Persistence**: localStorage key `tjsv.environmentMap` (`"true"`/`"false"`); Reset clears it and restores the page-load baseline (same layering as exposure/env-intensity/ambient/tone-mapping).
- **Resolution precedence** (unchanged pattern): URL `environment_map` param > `ThreeJSViewer` option > localStorage > hard default (`true`). A new `parseBool` accepts `true/false/1/0/on/off/yes/no`.
- **Python parity**: `ViewerClient(environment_map=True|False)` forwards `environment_map=true|false` as a query param — an authoritative initial value that wins over localStorage on reload (like the other lighting kwargs).

## Tests

- **Python** (`test_client.py`): `environment_map` serializes to `false`/`true`, and is omitted when unset.
- **Browser** (`test_browser.py`): the panel checkbox drops then restores `scene.environment` (map retained on `_envMap` throughout); the `environment_map=false` URL param starts with the env map off and the checkbox unchecked.

Full suite green (403 passed), `tsc --noEmit` clean, ruff clean, `viewer.html` regenerated (version placeholder intact).


tdamsma:
Addressed both Copilot comments in b0a5980:
- `client.py`: `environment_map` now validated as a real bool (raises `ValueError` on non-bool), so a stray `"false"` can't coerce to `environment_map=true`. Added a rejection test.
- `test_browser.py`: the toggle test now pins `environment_map=true` in the URL so a persisted `tjsv.environmentMap=false` can't make the initial assertion flaky.

PR #112, opened 2026-07-09
Accelerate polyline/tube picking on huge spines (#111)

### LINKED ISSUE #111: Accelerated polyline picking: _pickAt is O(N)-per-pointer-move over the full spine
The polyline/tube picker (`_polylinePick._pickAt`, called from `_onMove` on **every** `pointermove`) does a linear scan projecting **every** spine point to screen space (`_nodeScreenInto` → `applyMatrix4` per node). Cost is O(N) per mouse move, independent of render LOD.

## Impact
On a ~5M-point toolpath (native `THREE.Line`, one draw call — cheap to render), hover-picking dominates the frame budget: a Chrome profile shows ~63% of time in `_onMove → _pickAt → _nodeScreenInto → applyMatrix4`, hundreds of ms per move → abysmal framerate even at animation start, even in Line mode. The render is fine; picking is the killer.

## Ask (any one)
- **Spatial acceleration** for `_pickAt` (screen-space uniform grid / BVH over the spine, rebuilt on geometry or camera change), so a hover is ~O(log N) or O(cells).
- **Pick-stride / decimation option** on `enablePolylinePicking({ maxPickPoints })` — coarse nearest-segment on a strided spine, refined locally around the hit.
- **Hover throttle** — at minimum, coalesce `_onMove` picks to rAF (one pick per frame) so fast mouse moves do not queue multiple full scans.

## Current ribweaver workaround
`/sim` gates `enablePolylinePicking` **off** above 750k pickable points (`_PICK_MAX_POINTS`), trading pick→line sync for a smooth framerate on huge programs. Would drop the gate once accelerated picking lands.

Repro: load any >1M-point polyline with picking enabled, move the mouse over it, profile.

tdamsma:
Fixes #111.

## Problem
`PolylinePickController._pickAt` ran on **every** `pointermove` and did an O(N) linear scan projecting every spine node to screen space. On a ~5M-point toolpath (one cheap draw call to render) hover-picking dominated the frame budget — a Chrome profile showed ~63% of time in `_onMove → _pickAt → _nodeScreenInto → applyMatrix4`.

## Fix (two of the three asks in the issue)

**1. rAF hover coalescing — always on.** `_onMove` now only stashes the latest cursor position and defers the scan to `requestAnimationFrame` (`_onMoveRaf → _doHover`), so a fast mouse move that fires many `pointermove` events runs at most one full scan per frame.

**2. `max_pick_points` decimation — opt-in.** The inner scan is refactored into `_scanObjectSegments(..., step)`:
- **Coarse pass**: with `max_pick_points=N`, a spine with more than `N` nodes is visited with a stride so only ~`N` chord-segments are screen-tested.
- **Refine pass**: the winning object's coarse hit is re-scanned at full resolution (`step=1`) in a local window `[seg−stride, seg+2·stride]`.

Hover cost drops from O(N_spine) to O(N_pick). The resolved `segment`/`fraction`/`point` stay **exact** (refine runs on the true spine); only nearest-segment *selection* is approximate on sub-stride wiggles. `0`/`<2` = off (exact full scan, the default).

Lifts ribweaver's `/sim` 750k-point picking gate.

## Surface
- Python: `enable_polyline_picking(..., max_pick_points=0)` → `maxPickPoints` on the `set_polyline_picking` message.
- JS: `enablePolylinePicking({maxPickPoints})`, threaded through the message handler.
- `viewer.html` regenerated; CLAUDE.md picking section updated.

## Verification
- `npx tsc -p jsconfig.json` clean on the picking code; `ruff check`/`format` clean.
- Full suite: **408 passed**, plus 2 new protocol tests asserting the `maxPickPoints` wiring.


tdamsma:
Addressed the Copilot review comments in fdc9c11:

- **rAF hover after `pointerleave`** — added `_pointerOver` tracking so a coalesced hover scheduled via `requestAnimationFrame` doesn't resurrect the marker/cursor if the pointer already left the canvas before the frame ran.
- **Per-segment `Math.sqrt`** — `_scanObjectSegments` now compares squared pixel distances (`dist2`) in the hot per-move scan; the gate itself is squared once for comparison instead of sqrt-ing every segment's distance.
- **`max_pick_points` normalization** — `enable_polyline_picking` now normalizes values `< 2` to `0` on the Python side too (matching the viewer's own decimation-disable threshold), with a new protocol test pinning it.

Full suite (413 tests) passes.

Issue #119, opened 2026-07-13
strand_collapse: exempt rings on large segments from fold-snapping (and consider toning the snap down)


tdamsma:
Operator A/B on ribweaver /sim with `impression-cabine.mpf` (192k points, A45): the tuned default (`strandCollapse: true`, maxSnapFactor 1.0) produces corner-cut / wedge-hole artifacts — rings snapped up to ~12 mm off the true spine (8,269 rings >2 mm measured against the pre-collapse pickPoints). Disabling collapse entirely (`scSnap=0`) removes **100%** of the artifacts; a pipeline-side micro-segment merge alone removes ~95%. So the fold-detect-and-snap pass is over-reaching well beyond the tight wipe loops it exists for.

Requested tuning (keep the feature, constrain its reach):

1. **Exempt large segments**: a ring whose adjacent spine segments are long relative to the bead width (say ≥ 1 bead width) is on an open straight, not inside a fold — the snap pass should never move it, no matter what the fold detector claims (false triggers there come from degenerate tangents at nearby micro-segments/breaks, see #117). Collapse should only act within dense regions (consecutive short segments, the actual wipe-loop class).
2. **Tone the snap down**: consider a smaller default maxSnapFactor. Measured sweep at ribweaver-side: 0.5 → 4,877 rings >2 mm (still visible pinches); 0.25 → 222, visually clean at corners while still creasing genuine folds; 0.1 ≈ off. With the large-segment exemption in place the factor matters less, but 1.0 today moves geometry much further than a crease needs.

Interacts with #117 (breakMask-blind miter frames + the absolute 1e-12 near-duplicate guard): fixing those removes most of the false fold triggers; this issue is about bounding the blast radius when the detector fires anyway.

Repro data: ribweaver session artefacts (8,269/2,502 ring-deviation counts, maxSnapFactor sweep screenshots) — measured with the tube blob's pre-collapse spine vs rendered `geometry.position` on the live mesh.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01N8TR21cdEDzjpMsrfg7zJ8

tdamsma:
Fixed and shipped in release 0.0.44 (PR #120, commit a65f0a5). See the CHANGELOG 0.0.44 section for details.

Issue #118, opened 2026-07-13
Parametric-tube LOD not recomputed on close-in zoom (skip gate uses model size, not camera distance)


tdamsma:
`_dispatchLodWorker` skips a tube when the camera moved less than `0.05 * boundingRadius * 2` — an ABSOLUTE, model-sized threshold (the comment says '5% of distance to tube center' but the code never computes that distance). Zooming in close moves the camera by centimetres while the LOD epsilon (= cameraDist / LOD_EPSILON_DIVISOR) shrinks by an order of magnitude, so the rebuild never fires and a stale coarse tube renders up close (visible RDP faceting on ribweaver /sim programs in the 600k+ LOD regime).

Fix (verified locally, working-tree patch in this repo — yours to review/commit): reference distance = distance from the camera to the tube's bounding-sphere SURFACE (world-transformed), floored at 1% of boundingRadius, capped by the old 2×boundingRadius; threshold = 5% of that. Far away it behaves as before; close in, small camera moves re-trigger the worker.

Related bounded staleness that may remain by design: the worker's per-chunk reuse keeps cached RDP indices while the chunk's camera-distance ratio stays within LOD_CHUNK_REUSE_RATIO = 1.5, so detail refreshes in ~1.5× zoom steps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01N8TR21cdEDzjpMsrfg7zJ8

tdamsma:
Fixed and shipped in release 0.0.44 (PR #120, commit a65f0a5). See the CHANGELOG 0.0.44 section for details.

Issue #117, opened 2026-07-13
Parametric tube: zero-length spine segments snap the frame tangent to +X (fold-fan shards at bends)


tdamsma:
`computeMiterFrames` falls back to a hardcoded `(1,0,0)` tangent for zero-length segments (both the initial-seed block and the rolling interior update). A spine with exactly-coincident duplicate vertices — e.g. ribweaver /sim's #102 width-step twin rings — gets rings rotated to +X wherever the duplicate sits off an X-aligned run: invisible on X-parallel straights, fold-fan shards at bends.

Fix (verified locally, working-tree patch in this repo — yours to review/commit):
- initial seed: scan forward to the first non-degenerate segment instead of assuming segment 0;
- rolling update: on a zero-length segment carry the previous direction through (`c` keeps its value) instead of resetting to +X.

Both the main-thread `buildParametricTubeGeometry` and the LOD worker share `computeMiterFrames`, so one fix covers both. Repro: any binary tube whose spine contains coincident duplicates at a direction change; ribweaver `/sim` loading `impression-cabine.mpf` (A45 corpus) showed shards on every wipe-loop bend.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01N8TR21cdEDzjpMsrfg7zJ8

tdamsma:
Numeric trace on ribweaver /sim (impression-cabine.mpf, 192k points) sharpens this issue's scope — two compounding defects in `computeMiterFrames`:

1. **No `breakMask` awareness**: the bisector tangent + miter scale/direction are computed across consecutive indices even when a break (severed travel) sits between them. Break-adjacent rings deviate ~0.5–0.85× half-bead-width with NO micro-segment present (measured at all 13 breaks). Fix: thread the break mask in and force bevel (scale=1, no cross-boundary bisector) at severances — same spirit as the existing hairpin/cone handling.

[...]

PR #120, opened 2026-07-14
Parametric tube: frame tangents, LOD zoom gate, strand-collapse over-reach (#117, #118, #119)

tdamsma:
Three related parametric-tube quality fixes (issues #117, #118, #119).

## #117 — Zero-length spine segments no longer snap the frame tangent to +X
`computeMiterFrames` seeds from the first non-degenerate segment and carries the previous direction through a degenerate (exactly-coincident) one, instead of falling back to a hardcoded `(1,0,0)`. A duplicate vertex at a bend now renders as a clean step rather than a fold-fan shard. Shared by the main-thread build and the LOD worker.

## #118 — LOD-dispatch skip gate uses camera distance, not model size
The gate now references the camera's actual distance to the tube's bounding-sphere **surface** (radius scaled by the object's max world-axis scale; floored at 1% of the radius, capped by the old 2×radius proxy; no per-frame allocation), not an absolute model-sized threshold. Close-in zooming re-triggers the worker; previously the gate never fired up close and a stale coarse tube rendered with visible RDP faceting.

## #119 — strand_collapse: large-segment exemption **+ lower default snap factor**
Two changes, together:
1. **Large-segment exemption** — a ring whose shorter adjacent spine segment is ≥ `large_seg_factor` bead-widths (default 1.0) is treated as an open straight and never snapped, confining collapse to dense wipe-loop regions and removing false snaps from degenerate tangents at nearby micro-segments. Tunable via the `strand_collapse` dict (`0` = off).
2. **Default `max_snap_factor` 1.0 → 0.5.** The full-bead reach (1.0) over-snaps tight/messy folds — a spiky-cusp fin + z-fighting seam at the crease, and over-snap spikes shooting out along the adjacent straight on jittered reversals. Going all the way down to 0.25 under-shoots real wide-bead corners instead (strands stop short of the apex → protruding wedge vs the #50 ribweaver-bulb tuned baseline). **0.5 is the sweet spot**: it reaches the apex identically to 1.0 on real corners (0.00% pixel diff vs the #50 bulb baseline) while staying gentle enough to avoid the tight-fold artefacts. See the snap-factor comparison images in the comments (the thread documents the tuning iteration; 0.5 is what ships).

Also fixes a **latent bug**: the non-LOD tube path reconstructed the collapse config as `{maxSnapFactor}` only, silently dropping `largeSegFactor` so the knob never reached the collapse pass.

### Validation
Headless harness reading deviation + quality measures straight off the WebGL buffer (`tmp/tube_lab`, moving to `bench/tube_lab` in a follow-up):
- **Snap 0.5 across regimes**: tight hairpin (no cusp fin, unlike 1.0), the real #50 bulb apexes (0.00% pixel diff vs the tuned 1.0 baseline), a sharp jittered ±120° reversal (no over-snap spikes), a rounded jittered S-curve. Dense-hairpin displacement drops from 7.9 mm max (168 rings moved) at 1.0 to 4.0 mm max (85 rings) at 0.5.
- **Quality measures on the realistic decimated+jittered ±120° zigzag**: 0.5 cuts intersecting-wireframe-edge pixels to **76%** of collapse-off at footprint **IoU 0.998**, unchanged triangle count — and scores identical to 1.0 (the default sits at its plateau; nothing above 0.5 buys more).
- The large-segment exemption is inert on genuine dense folds (all segments ≪ bead width, nothing qualifies) and only ever *skips* a snap, never harms a crease.

## Tests
- 4 serialization tests for `large_seg_factor` (valid / zero-kept / negative-rejected / non-number-rejected).
- Existing strand_collapse browser regression tests pass at the new default (genuine fold fires, ≤ max(W,H) displacement, toggle round-trips).
- Full suite: 418 unit + tube/toolpath browser tests green; `tsc` clean.


tdamsma:
### #119 snap-factor comparison — dense hairpin (genuine fold)

Left→right: `strand_collapse=off`, `before` (exemption off), `after` (default, exemption on), snap `0.5`, snap `0.25`. The `after`/default column is **identical** to `before` (the exemption is inert on a genuine dense fold), while snap `0.25` visibly **under-creases** the fold — evidence for keeping `max_snap_factor=1.0`.

hairpin snap factors

tdamsma:
> ### #119 snap-factor comparison — dense hairpin (genuine fold)
> 
> Left→right: `strand_collapse=off`, `before` (exemption off), `after` (default, exemption on), snap `0.5`, snap `0.25`. The `after`/default column is **identical** to `before` (the exemption is inert on a genuine dense fold), while snap `0.25` visibly **under-creases** the fold — evidence for keeping `max_snap_factor=1.0`.
> 
> hairpin snap factors

from left to right 1 looks ok, 2 and 3 are broken, horrible artefacts. , 4 and 5 look ok, 6 looks bad as the corner is compressed too much

tdamsma:
You're right — I misread my own metric. The jagged sawtooth + the fin spiking out of the cusp in columns 2–3 (snap **1.0**) are over-snap artefacts, not a "fuller crease"; my deviation number equated *more ring movement* with *better*, which is backwards on a tight fold. Columns 4–5 (0.5, 0.25) are the clean ones — exactly as your issue #119 measurement already found (0.25 clean, 0.5 still pinches).

**Fixed** in `c9a73c2`: default `max_snap_factor` **1.0 → 0.25**. On the same dense hairpin, the shipped default now moves rings ≤2mm (was 7.9mm) and the cusp fin is gone:

left→right: `off`, `snap 1.0` (OLD default — fin + sawtooth), **`default` (new 0.25 — clean cusp)**, `snap 0.5`, `snap 0.25`.

default 0.25 vs 1.0

Also addressed both Copilot review points in the same commit:
- **LOD gate** now scales the bounding-sphere radius by the object's max world-axis scale (was using the unscaled local radius against a world-transformed center — wrong for a scaled tube) and reuses module-level scratch vectors instead of allocating a `Vector3` per frame.
- **`segLen`** allocation + per-segment sqrt is now skipped entirely when the exemption is off (`large_seg_factor=0`).

The genuine-fold browser regression test still passes at the new default (fold fires, ≤ max(W,H) displacement). PR #121 rebased on top.

tdamsma:
### Regression check — the 0.25 default vs the tuned strand_collapse cases (#50/#51)

Concern: lowering `max_snap_factor` 1.0→0.25 could under-crease the wide-bead-corner cube that #50 tuned at 1.0. Reproduced the exact #50 scenario — the 100-pt ribweaver bead (`tube_8f5bba97`), via `tools/capture_strand_collapse.py` — at collapse `off` / `0.25` (new default) / `1.0` (old/tuned):

wide-bead apex off/0.25/1.0

**No regression.** Pixel diff of `0.25` vs the tuned `1.0` on the tuned sites:

| site | 0.25 vs 1.0 (tuned) | off vs 1.0 |
|---|---|---|
| bulb apex A (spine 28) | 0.2% px | 0.7% px |
| bulb apex B (spine 58) | 0.2% px | — |
| rectangle corner (spine 12) | 0.1% px | — |

The collapse still fully creases the wide-bead apex (off→1.0 changes 0.7% of pixels; the crease fan is intact) and **0.25 reaches the same apex as 1.0** — on this bead the fold distance is under 0.25·max(W,H), so both cap identically. The `0.25`↔`1.0` difference only shows up on an *extreme* synthetic fold (radius-2 bead-8 hairpin, ratio 4×) where 1.0 over-snapped into the cusp fin; real wipe loops (ratio ~1–2) collapse the same at either factor.

Separately, #56's corner geometry (miter / frame-freeze / bias) is not touched by the snap factor and is protected by its 7 regression tests — all green. Today's build at `1.0` also matches the #50 AFTER baseline, confirming #117/#118 didn't perturb the collapse geometry.

Issue #113, opened 2026-07-09
Parametric tube silently truncates past ~833k ring-pairs: single draw exceeds WebGL max index count (30M)


tdamsma:
A large single `add_parametric_tube_binary` builds one BufferGeometry whose **index count exceeds the WebGL context per-draw limit**, so the GPU draws only the first N indices and the rest of the tube is **invisible** — no error, just a `drawElementsInstanced` warning.

## Repro
A ~1.24M-spine tube (1,238,919 ring-pairs) → geometry with **44,601,660 indices** (Uint32). On Firefox/ANGLE:

```
WebGL warning: drawElementsInstanced: Context's max indexCount is 30000000,
but 44601660 requested. [webgl.max-vert-ids-per-draw]
```

The tube renders correctly up to **30,000,000 ÷ 36 = 833,333 ring-pairs**, then vanishes for the rest. (36 = nCs(6) × 2 tri × 3.) `geometry.index.count` reports the full 44.6M and `drawRange.count` is null (all), so nothing in the JS state hints at the truncation — it's a hard GPU cap. The limit is driver-dependent (30M here; other drivers/`webgl.max-vert-ids-per-draw` differ), so it can't be assumed safe.

## Impact
Any tube past ~833k ring-pairs is partially invisible in bead view. Consumers hit this on large toolpaths (ribweaver `/sim` renders whole post-processed MPFs — 1M+ points is normal). LOD (`lod:true`) mitigates the zoomed-out view but decimates by camera distance, so a close zoom on a big tube can still exceed the cap.

## Ask
Split a tube whose index count would exceed a conservative cap into **multiple draw calls** — either geometry groups / multiple `drawRange` sub-draws, or multiple sub-meshes — each under the cap (e.g. min(driver `MAX_ELEMENT_INDEX`-ish / queryable limit, a safe 16–24M default). Draw-range reveal (`applyParametricTubeDrawRange`) then needs to span the chunks. Alternatively expose the effective cap so callers can pre-chunk.

Non-LOD path is where ribweaver `/sim` hits it (it sends `lod:false`).

tdamsma:
Fixed and shipped in release 0.0.44 (PR #120, commit a65f0a5). See the CHANGELOG 0.0.44 section for details.

Issue #114, opened 2026-07-09
LOD tube should cap refinement to the WebGL per-draw index limit (not just camera-distance)


tdamsma:
Follow-up to #113 (low priority). Enabling `lod:true` on a large parametric tube fixes the zoomed-out view, but LOD decimation is driven purely by camera-distance chord tolerance (epsilon = distance / DIVISOR) with **no hard index-count budget**. Zoom in close on a very large tube and LOD refines globally toward full resolution, so the rendered index count can again exceed the WebGL per-draw cap (~30M idx on Firefox/ANGLE, `webgl.max-vert-ids-per-draw`) and the tail silently truncates once more.

## Ask
Clamp LOD level selection so the rendered index count never exceeds a conservative per-draw cap (queryable limit, or a safe 16–24M default) — i.e. treat the WebGL index cap as a hard ceiling on top of the distance-based tolerance. Once #113 (chunk the tube into multiple sub-draws) lands this is subsumed, but as an interim it makes `lod:true` a complete fix rather than a zoom-dependent one.

Repro: 1.24M-spine tube (44.6M idx), `lod:true`, zoom in close → tail disappears again. See #113 for the full numbers.

tdamsma:
Fixed and shipped in release 0.0.44 (PR #120, commit a65f0a5). See the CHANGELOG 0.0.44 section for details.

PR #121, opened 2026-07-14
Parametric tube: split huge tubes across the WebGL per-draw index cap (#113, #114)

tdamsma:
Fixes issues #113 and #114. **Stacked on #120** (base branch `fix/tube-frame-collapse`) — review/merge that first; the net diff here is only the index-cap code.

## Problem
A tube whose index buffer exceeds the WebGL per-draw limit (`webgl.max-vert-ids-per-draw`, 30M on Firefox/ANGLE; driver-dependent) is **silently truncated** by the GPU — the tail of a >833k-ring-pair tube vanishes with only a `drawElementsInstanced` console warning. Consumers hit this on large toolpaths (a 1.24M-spine tube is 44.6M indices). `lod=true` only mitigated the zoomed-out view; a close zoom refined back over the cap (#114).

## Fix
`applyTubeDrawCap` splits any tube over a conservative `MAX_TUBE_INDICES_PER_DRAW = 24M` into `geometry.addGroup` chunks (whole-triangle-aligned, ≤ cap) + a 1-element material array, so three.js issues one `drawElements` per group under the cap. Tubes under the cap keep a single material and no groups — **byte-identical to before**.

Groups partition the index buffer **positionally**, and the r183 renderer intersects `geometry.drawRange` with each group's `[start, start+count)`. So the existing single `setDrawRange(0, N)` reveal and the endCap relocation span chunks with **no special handling** — verified in a browser test. Applied on both the non-LOD build and every LOD-worker geometry swap, so an LOD tube refining past the cap up close is chunked too instead of truncating again — **subsuming #114**. The M/N debug material swap is group-aware so a chunked tube doesn't truncate in debug view.

## Verification
- 3 browser tests via a `window.__tubeDrawCapOverride` test seam (lowered cap, so we don't build a real 24M-index tube): groups tile `[0, total)` contiguously, each triangle-aligned and ≤ cap; `draw_range` spans chunks; under-cap tubes stay single-material with no groups.
- Visual: a forced-chunked helix (3 groups, material array) renders with correct shading, vertex colors, end cap, and a `draw_range 0.6` reveal that crosses all 3 chunk boundaries seamlessly — no regression from the material-array path.
- Full tube+toolpath suite (84 tests) green; `tsc` clean.


tdamsma:
### Chunked tube renders correctly (forced 3-group split, `draw_range 0.6`)

A helix tube forced over a lowered cap into **3 geometry groups + a material array** (14,940 indices). Shading, vertex colors, and the end cap are intact, and the `draw_range 0.6` reveal crosses all three chunk boundaries seamlessly — no regression from the material-array path. (A true before/after of the *truncation* bug would need a real >30M-index tube; at test scale both look identical, which is the point — chunking is transparent.)

chunked tube draw range

PR #127, opened 2026-07-21
Add add_grid: shader floor-grid primitive (anti-aliased, distance-faded infinite grid)

### LINKED ISSUE #126: Shader floor-grid primitive (anti-aliased, distance-faded infinite grid)
Follow-up from CEAD-group/ribweaver#505 (CSD additional rendering).

The /cells lane now renders a floor grid server-side, but the current wire protocol can only express a **line grid** (`add_polyline_binary` segments) or `show_grid` (`THREE.GridHelper`). The reference `geometry-loader.js` prototype uses a **ShaderMaterial** floor plane: crisp anti-aliased grid lines, a distinct center-line/edge colour, and screen-space-stable line width that reads well at grazing angles and large extents — none of which a raw `LineSegments` grid gives.

Ask: a first-class grid primitive (e.g. `add_grid` / a grid material on a plane) that takes cell size, extent, line width, and center/edge/background colours, rendered with the prototype's shader look. Until then ribweaver ships the line-grid fallback.

No API break needed — purely additive.

tdamsma:
Closes #126.

Adds `add_grid(id, cell_size, extent, line_width, color, center_color, background_color, background_opacity, fade_start)` — a first-class shader floor grid with the look of ribweaver's `geometry-loader.js` prototype, improved where the prototype was fixed-function:

- **Screen-space-stable anti-aliased lines**: line distances are measured in pixels via `fwidth`, so `line_width` (px) holds at any zoom and grazing angle, with a 1px smoothstep AA edge (the prototype's UV-space `smoothstep` width varied with plane size).
- **Cell-density thinning**: where a cell spans under ~4 px the grid fades out instead of collapsing into a solid sheet.
- **Distinct centre/axis lines**: the two lines through the local origin draw in `center_color`, slightly wider.
- **Distance fade**: radial alpha fade from `fade_start`×half-extent to the plane edge, so the finite plane reads as an infinite floor (alpha fade rather than the prototype's colour-mix-to-background, so it composites correctly over any content).
- **Optional fill**: `background_color` + `background_opacity` tint the plane between the lines.

Behaviour: tracked in `_objects` (id, delete, visibility, transform, parent, dropped by `clear`; multiple grids coexist — see the fine+major pair in `examples/29_floor_grid.py`), but **excluded from scene bounds** (a 10 km floor must not inflate framing/near-far), **never raycast** (picks pass through to geometry behind), and skipped by the `M`/`N` debug material cycles. The fragment shader ends with the `tonemapping_fragment`/`colorspace_fragment` chunks so it follows renderer tone mapping like built-in materials, including under the EDL composer. `show_grid` (GridHelper) is untouched. Purely additive — no API break.

Tests: protocol payload + validation unit tests, and a browser test asserting the grid appears, is excluded from scene bounds, and deletes cleanly. Full suite (433, incl. Playwright) passes; `tsc --noEmit` and ruff clean.

Screenshots incoming in a comment.


tdamsma:
### Screenshots (`examples/29_floor_grid.py`: 1 m grid + 10 m major grid + axis lines)

**Overview** — AA line quality, brighter axis lines through the origin, major/minor pair:
overview

**Grazing angle** — cell-density thinning + distance fade: the grid dissolves toward the horizon instead of collapsing into a solid sheet:
grazing

**Top-down** — uniform 1.5 px lines across the whole plane (screen-space-stable width):
top

tdamsma:
Re the Copilot review:

- **`fade_start=1.0` / smoothstep equal edges** — valid, fixed: `uFadeStart` is now capped at 0.999 in `buildFloorGridMesh` (GLSL leaves `smoothstep` undefined for `edge0 >= edge1`).
- **`extensions: { derivatives: true }`** — not applicable: the viewer ships three r183, which is WebGL2-only (`fwidth` is core GLSL ES 3.00; the `extensions.derivatives` flag no longer exists in three).
- **truthiness on `position`/`rotation`/`scale`** — not a bug: `[0, 0, 0]` is a non-empty list and therefore truthy in Python, so it is sent; only `None`/empty are skipped. This also matches the existing `_add_primitive` convention used by every other `add_*` method.

tdamsma:
Second Copilot round addressed in the latest commit:

- **Hidden grid inflating the near/far fit** — valid, fixed: invisible `excludeFromBounds` objects are now skipped when building the near/far sphere, so hiding a grid restores full depth-buffer precision.
- **Example comment mismatches** ("400 m" vs `extent=40.0`, "slightly above" vs `-0.001`) — the values are the intentionally tuned ones; the stale comments were updated to match (40 m fine plane; the major grid sits just *below* the fine one so the two transparent planes never z-fight).

PR #160, opened 2026-08-03
Slim translate gizmo pickers so arrows stop shadowing the plane chips

tdamsma:
## Problem

Careful mousing on a translate plane chip (e.g. the blue XY parallelogram) often grabbed an axis arrow instead. TransformControls resolves hover closest-intersection-wins across all pickers, and the stock arrow pickers are fat cone volumes hugging each axis (radius 0.2, vs 0.04 for the visible cone). From any 3/4 view those volumes bulge in front of the flat plane chips, so the arrow was genuinely nearer along the view ray over roughly half of each chip's visible area — even though the chip's own picker is deliberately larger than its visual (`GIZMO_PLANE_SCALE`). Aggravated since `GIZMO_SIZE = 0.5` (#155) packed the same stock picker volumes into half the pixels.

## Fix

`refineGizmoHandles` (the one-time structural pass, shared with the clip gizmos) now also slims the translate pickers:

- each arrow picker's two radial (non-axis) geometry dims scale by `GIZMO_ARROW_PICKER_SLIM = 0.4` (radius 0.2 → 0.08, still ~2× the visible cone's hit slop);
- the centre free-move `XYZ` octahedron scales by `GIZMO_CENTER_PICKER_SCALE = 0.6`.

The handle's axis passes through the baked geometry's radial origin, so a plain `geometry.scale` keeps the picker centred on its axis; guarded one-time via `userData.__slimPicker`.

## Verification

Synthetic pointer-move grid scan reading `control.axis` around the XY chip (`#` = XY chip hit, `x` = X arrow, `*` = centre, `3` = XZ chip, `o` = inside the visible chip but nothing hit):

**Before** — the X arrow claims the middle of the blue parallelogram:

```
xx*****333333######oo
xx*******33333#######
yy******xxxxxxxxx####
yyy****xxxxxxxxxxx#oo
yyy*****xxxxxxxxxxxxo
```

**After** — the chip's hover area covers its parallelogram (`o` cells are bbox-vs-parallelogram corner artifacts, not dead zones):

```
xx...z.3333333####ooo
..x****3333333######o
...*****333333########
...*****..3333#######
..y.***.x.oo33######o
```

Arrows remain hittable along their length and tips (probed all 6 arrow picker midpoints). Type-check clean; 22 gizmo browser tests + 346 unit tests pass.

Before/after screenshots of the picker volumes and the hover highlight follow in a comment.


tdamsma:
Comment by Claude (bot)

Before/after visuals (same scene, same camera, cursor at the **same client pixel** inside the blue XY chip — a point the old build resolves to the X arrow):

### Hover at the same pixel on the chip

| Before — X arrow steals the hover (`control.axis = "X"`) | After — chip wins (`control.axis = "XY"`) |
|---|---|
| before hover | after hover |

### Picker (hit-test) volumes made visible (orange = arrows/centre, magenta = plane chips)

| Before — fat stock arrow pickers bulge in front of the chips | After — slim spokes, chips reachable |
|---|---|
| before pickers | after pickers |

Captured with a Playwright script that loads each built `viewer.html` standalone (`handleMessage` direct feed, no WS), picks the steal point automatically by scanning `control.axis` over the chip's projected bbox, and reuses the identical pixel for both builds.

tdamsma (commented):
Comment by Claude (bot)

Review of the diff (viewer.js + regenerated viewer.html + CLAUDE.md):

**Correctness**
- The radial `geometry.scale` is valid for these specific pickers: TransformControls bakes each handle's offset into its geometry, and every translate arrow picker's axis passes through the radial origin of that baked geometry (verified live on both `+`/`-` picker meshes per axis), so scaling the two non-axis dims keeps the picker centred on its axis. The `XYZ` octahedron is origin-centred, so uniform scale is exact.
- `userData.__slimPicker` makes the pass idempotent; chips fall through `else continue` untouched (their enlargement stays owned by `restyleGizmoHelper`'s `sizedPlanes` path — no double-ownership).
- `computeBoundingBox`/`computeBoundingSphere` after scaling keeps the raycaster's sphere pre-test tight — correct and cheap (one-time).
- Rotate-mode pickers untouched (no chips compete there); `scale` mode is not exposed (`_ALLOWED_GIZMO_MODES = {"translate", "rotate"}`), so stock scale pickers are unreachable.
- Shared-pass reach: the clip tool's Z-only slide gizmo inherits a slimmer Z picker. It has no competing chips, keeps ~2× the visible cone's hit slop, and `test_clip_tool_has_refined_rotate_and_slide_gizmos` passes — acceptable.

**Conventions** — constants + banner-comment style match the file; `viewer.html` regenerated via `build.py`; CLAUDE.md updated as the living doc; `npx tsc --noEmit` clean.

**Verification** — 22 gizmo browser tests + 346 unit tests pass; the PR body's before/after `control.axis` grid scans and the screenshot comment demonstrate the fix and that arrows/tips remain hittable.

**Finding (should fix before merge)**
- *Test coverage*: nothing pins the fixed behavior — a future three.js bump (new stock picker shapes) or picker tweak could silently reintroduce the arrow-steal. Suggest a browser test that sets a deterministic camera pose, projects the XY chip's bbox, dispatches a `pointermove` grid over it, and asserts the chip wins the majority of resolved hits (and that each arrow still resolves somewhere along its shaft). This encodes the actual UX contract rather than the implementation (picker sizes).

Otherwise LGTM.

tdamsma:
Comment by Claude (bot)

Review finding addressed in ee66325: added `test_gizmo_plane_chip_hover_beats_arrow_pickers` — a pointermove grid scan over the XY chip's projected bbox from a 3/4 view asserting the chip wins its own footprint, plus per-arrow probes asserting every arrow still resolves at its shaft/cone centres. Verified red against the pre-fix `viewer.html` (the X arrow claimed the chip) and green with the fix; full gizmo suite passes (23).

PR #168, opened 2026-09-01
Orbit-pivot fallback ignores invisible objects (#166)

### LINKED ISSUE #166: Orbit center includes invisible meshes
_Filed by Claude (bot) on behalf of Thijs._

When orbiting, the rotation center now seems to be computed from all meshes, visible and invisible alike. It should be computed from the visible geometry only — hidden objects should not pull the orbit pivot.

tdamsma:
Fixes #166.

## Problem
The orbit-pivot fallback (a click that hits no component) took the center of `updateSceneBounds()`'s content sphere. That sphere deliberately **does not** filter on visibility — a hidden object may be shown later and must not end up outside the fitted near/far planes — so any hidden mesh far from the model dragged the orbit center off-screen.

## Fix
The fallback now reads `_collectFrameableBounds()`, the same AABB Home/F frame to: it bails on `.visible === false` subtrees and already skips both grid kinds, the clip/move gizmo helpers, and the pivot marker. Near/far fitting is untouched, so a hidden object still can't be clipped when it later becomes visible.

## Test
`test_orbit_pivot_ignores_invisible_objects` — a visible box at x=10 and a hidden one at x=1000: the pivot lands on x≈10 while `_nearFarSphere` still reaches the hidden one. Verified it fails on `main` and passes with the fix.


tdamsma:
Comment by Claude (bot)

Rebased onto `main`. This branch had been cut from a local `main` that carried three unpushed `append_points` commits, so the PR diff showed that whole feature on top of the actual fix. Those commits are now their own PR (#174) and this branch contains only the orbit-pivot change: 4 files, ~55 lines.

That resolves the Copilot review note about the PR's scope. The other Copilot note (`clear()` not evicting `_append_blob_keys`/append blobs) belongs to the `append_points` code and has moved to #174.

Issue #145, opened 2026-07-23
Sanctioned unknown-message hook on the viewer WS (retire ribweaver /sim onmessage monkey-patch)


tdamsma:
🤖 bot (filed by overnight triage on Thijs's behalf)

**Ask**: a sanctioned hook for application-defined message types on the viewer WS, e.g. `viewer.onUnknownMessage(cb)` — invoked from `handleMessage`'s `default:` branch (viewer.js ~line 11636) with the parsed `data`, replacing (or before) the `console.warn`. Return-value or registration semantics are the implementer's call; a single callback is enough.

**Why**: ribweaver's `/sim` lane multiplexes app frames (`{type:'artifact'|'reinterpret_done'|'reinterpret_error'|'ingest_progress'}`) over the same WS the viewer owns. Today it monkey-patches `viewer3d._ws.onmessage` with a retry loop (`wrapArtifactIntercept`, `ribweaver/web/sim/static/sim.html` ~line 871) — brittle against reconnects and viewer internals. CEAD-group/ribweaver#408 tracks retiring that patch once a sanctioned hook ships; a viewer-side sim.html comment already (incorrectly) claims an `onAnyMessage` exists, so the name `onUnknownMessage`/`onAnyMessage` is bikesheddable.

Note the hook should also fire for unknown types arriving via a direct `handleMessage()` call (the embedder path), which the current monkey-patch misses.

PR #169, opened 2026-09-01
Embedder load-progress / assets-loaded hooks (#163)

### LINKED ISSUE #163: Expose an embedder load-progress/completion callback for asset fetches
*(Posted by an AI agent on behalf of Thijs Damsma.)*

The viewer knows exactly when the assets it fetched are complete (`_maybeNotifyAssetsLoaded`) but reports it only to the server (`assets_loaded`). An embedding page that wants a loading overlay has to poll the private `_pendingFetches` counter — which ribweaver's /calib lane now does, with a fallback deadline for when that private name changes.

Ask: an embedder callback — `onAssetsLoaded()` and/or `onLoadProgress(loaded, total)` — so lanes can show honest load state without reaching into private viewer internals.

tdamsma:
Closes #163.

An embedding page that wants a loading overlay had to poll the private `_pendingFetches` counter (ribweaver's /calib lane does, with a deadline fallback for when the private name changes). This adds the public surface.

## API (JS-only — no Python/WS surface)

- `viewer.getLoadState()` → `{loaded, total, pending, assetsComplete}`
- `viewer.onLoadProgress(cb)` — fires on every asset-fetch **start** and **end** with that snapshot
- `viewer.onAssetsLoaded(cb)` — fires when the viewer goes fully idle

Both `on*` return an unsubscribe fn, per the house hook idiom (`onUnknownMessage`, `onAnimationTime`, `onPolylineHover`).

## Why not a real `total`

Binary payloads are fetched as their messages arrive, so the viewer genuinely cannot know an up-front total — inventing one would be a lie. The counters describe the current **load batch** instead:

- a batch opens when `_pendingFetches` goes 0 → 1 and closes when it returns to 0;
- `total` = fetches **started** in that batch — a running count of what has been *seen*, which may grow mid-batch;
- `loaded` = how many of them finished;
- the finished batch's counters are retained while idle and reset by the next batch's first fetch, so the last progress event stays valid.

`pending === 0` is the honest "spinner off" signal: everything the viewer knows about has landed. It does not promise the producer has no more messages coming — that is `onAssetsLoaded`.

## The no-WebSocket gap

`onAssetsLoaded` is the browser-side twin of the existing `assets_loaded` WS notification: it fires when the batch drains **and** the producer declared its stream complete via `mark_assets_complete` (Python's `wait_for_assets()` sends it; a no-WS embedder feeds it through `handleMessage()` after its last add). `_maybeNotifyAssetsLoaded` still gates the *socket send* on an open WS, but the hooks are ungated — the old WS-OPEN early-return locked out exactly the direct-`handleMessage()` embedder this issue is about.

## Tests

`test_load_progress_and_assets_loaded_hooks` drives two 404 blob fetches with `viewer._ws` detached, asserts the four progress events `(0,1,1) (0,2,2) (1,2,1) (2,2,0)`, that no completion fires before `mark_assets_complete`, that it fires with no socket after it, and that both unsubscribes stop delivery.

Full suite green except the pre-existing `test_set_view_cancels_drag_inertia` failure (fails on `main` too). `npx tsc --noEmit -p jsconfig.json` clean, ruff clean, `viewer.html` regenerated.


tdamsma:
Comment by Claude (bot)

I dug into the review note about repeat firing, and I'd push back on it — with a test to make the reason stick.

**Applied to the WS message it would be a bug.** `assets_loaded` is a *reply*, not an event: `wait_for_assets()` clears its event, sends `mark_assets_complete`, and blocks until the viewer answers. A script that stages its scene in two passes calls it twice with no new assets in between, and suppressing the second reply on "nothing changed since the last one" reasoning hangs it until the timeout. Added `test_wait_for_assets_replies_to_every_marker` driving exactly that: two passes with nothing between them, then a third after real work.

**For the hooks it would trade information for nothing.** Every fire reports a true state — idle, with the producer's stream marked complete — and a repeat answers the same question the same way. Hiding an already-hidden overlay costs nothing, and `onAssetsLoaded`'s JSDoc already documents that it may fire again when further assets arrive. Gating on a batch generation would suppress the genuine "this batch finished too" signal to remove a duplicate that costs nothing.

Both decisions are now written down at the call site so the next reader doesn't re-derive them.

Also, on your question of why this exists at all: issue #163 is concrete about it — ribweaver's /calib lane currently polls the **private** `_pendingFetches` counter, with a deadline fallback for when that private name changes out from under it. This is the same shape as `onUnknownMessage` (#145) retiring the `ws.onmessage` monkey-patch: an embedder is already reaching into viewer internals to do something reasonable, and this gives it a supported surface. The part that isn't just wrapping existing state is that `_maybeNotifyAssetsLoaded` was gated on an open WebSocket, so a no-socket page feeding `handleMessage()` directly could never learn it was done — that gate is what the PR lifts.

Issue #147, opened 2026-07-28
Add a highlight/outline toggle for a mesh (set_highlight)


tdamsma:
> 🤖 Filed by an agent on Thijs's behalf.

## Request

A **highlight / outline toggle for a mesh** in the wire protocol — turn a persistent visual highlight on and off for a given object id.

## Why

ribweaver's `/cells` editor (the CSD cell editor) has one selection slice that currently projects to three places: the outline list, the Form view's component card, and the placement gizmo. Thijs wants selection to mean the same thing everywhere, including in 3-D — *"selecting a component in the form view and in the 3d view should be the same … item is selected, can it be outlined in 3d?"*

Today there is no way to say "this object is the selected one". Ribweaver can already draw parent/own coordinate frames for a selected component (ribweaver `0366517c`), and a *placeable* component gets a transform gizmo — but a robot link (J1…J6) gets neither a gizmo nor any visual mark, so on a 44-component cell there is no feedback at all that a given link is selected. An outline is the missing primitive, and it's the one that works uniformly for every component regardless of whether it's placeable.

## Current state (as of `ed7929d`, 0.0.46)

- No highlight/outline verb exists. The `switch` on message `type` in `src/threejs_viewer/viewer/viewer.js` covers `add_object`, `set_color`, `set_opacity`, `set_move_gizmo`, `batch_update`, `delete_object`, … but nothing for selection state.
- `set_color` is not a substitute: it destroys the mesh's authored appearance, and restoring it means the caller has to remember and re-push the original — which for a GLB-loaded CAD mesh with per-material colours isn't recoverable.
- There *is* existing outline machinery for the gizmo's own handles — `__gizmoOutline` / `__hasOutline`, around `viewer.js:6244-6305` — which may be the idiomatic in-repo starting point.

## Suggested shape

Mirror the neighbouring setters (`set_color` / `set_opacity`): object id + value, tolerant of an unknown id.

```
{type: 'set_highlight', id: '<object id>', enabled: true|false, color?: '<css colour>'}
```

with the matching Python sender alongside the existing ones.

## Requirements that matter for this use case

- **Toggle, not one-shot** — selection moves between components constantly.
- **Works on GLB/model-loaded meshes**, since that is what a CSD component is.
- **Idempotent**: setting twice must not stack two outlines.
[...]

PR #170, opened 2026-09-01
set_highlight: silhouette contour by default, tunable edge outline (#165, #158)

### LINKED ISSUE #158: set_highlight on dense CAD meshes reads as X-ray wireframe — expose EdgesGeometry threshold (follow-up to #147)
> Filed by Claude (bot) on Thijs's behalf, follow-up from #147 / CEAD-group/ribweaver#640.

## Observation

`set_highlight` (v0.0.47) builds its outline from `EdgesGeometry` with the default threshold angle. On prismatic/simple bodies (primitives, the extruder body) it reads as a clean selection outline — but on a **dense CAD mesh** (robot links J1..J6 in ribweaver /cells) the outline keeps so many interior edges it reads as an X-ray wireframe overlay rather than a silhouette. Same effect on curved analytic surfaces (a highlighted sphere shows its full triangulation grid). Screenshots: the J3 capture on the ribweaver `pr-640-images` pre-release, and the sphere capture on `pr-152-images`.

## Ask

Make the highlight legible on dense meshes. Candidate knobs, roughly in ascending effort:

1. **Expose the `EdgesGeometry` threshold angle** on the verb (`{type:'set_highlight', threshold_deg?: number}`, default 1°) — a higher threshold (e.g. 30–45°) drops interior/tessellation edges and keeps sharp features. Cheap, no new rendering machinery, probably sufficient for CAD links.
2. **Depth-tested outline variant** — draw the edge overlay with `depthTest: true` so only front-facing edges show (loses the always-on-top x-ray-through-occluders property, which may actually be desirable).
3. **True silhouette outline** (OutlinePass-style post-process or inverted-hull) — a real contour regardless of tessellation; heaviest option, interacts with the EDL composer.

Option 1 seems like the right first step; 2 could be a second boolean. The verb's reversibility/idempotence contract must hold for whatever ships (re-send with a new threshold re-tints/rebuilds in place).

### LINKED ISSUE #165: Mesh highlight: outline instead of full wireframe
_Filed by Claude (bot) on behalf of Thijs._

Highlighting a mesh currently renders a full wireframe over it. A silhouette/outline highlight would be much nicer — less visual noise, especially on dense meshes.

tdamsma:
Fixes #165. Fixes #158.

## Problem
The #147 highlight drew `EdgesGeometry` at THREE's own 1° threshold, so on a dense CAD mesh (ribweaver's robot links) or any curved analytic surface it kept nearly every interior edge and read as an X-ray wireframe overlay instead of a selection outline.

Raising the threshold alone (issue #158's option 1) is not enough: a smooth sphere has **no** crease above 30° at all, so the highlight would simply disappear on smooth bodies. So this ships both.

## What shipped
`set_highlight(id, enabled=True, color=None, style=None, width_px=None, threshold_deg=None)`

- **`style="silhouette"` (new default)** — a true contour: the same geometry drawn again with `side: BackSide`, each vertex pushed outward along its projected normal by `width_px` (default 3) **screen pixels** in clip space, scaled by `gl_Position.w`, so the rim is a constant pixel width at any zoom and under both cameras. The hull's backfaces fail the depth test wherever the object covers, leaving only the rim. It **shares the mesh's geometry** (no duplicate buffers — so disable disposes only its material), and `width_px` changes are a uniform write, not a shader recompile. The viewport uniform is `viewer._highlightResolution`, shared by reference into every silhouette program and updated in `resize()`.
  Deliberate trade-off: the hull must be depth-tested (a `depthTest:false` hull would paint its backfaces over the object body), so a silhouette highlight is hidden by objects in front of it, where the edge outline is always-on-top. #158 notes this may in fact be desirable.
- **`style="edges"`** — the original always-on-top feature-edge outline, now with `threshold_deg` (default raised 1° → 30°). Use it when the interior structure of the selection matters, or when the selection must show through occluders.
- A **skinned mesh**, or one whose geometry has no `normal` attribute, silently falls back to `edges` (the hull's basic material is un-skinned and extrudes along the normal).

The `M`/`N` debug traversals now skip highlight outlines — the hull *is* a Mesh, so a wireframe or debug-material swap would turn the contour into a solid shell. Idempotence and full reversibility are preserved: a re-send rebuilds only what changed (new threshold → swap EdgesGeometry; new width → uniform write; new style → dispose and rebuild), never a second outline, and the mesh's own materials are still never touched.

## Tests
`tests/test_browser.py`: default style/material state (BackSide, depth-tested, shared geometry, live viewport uniform), an **end-to-end pixel test** that the silhouette actually renders a rim around a smooth sphere and leaves nothing behind on disable, the edges-threshold behaviour (sphere: 0 segments at 30° vs the full wire grid at 1°; box: exactly 12 edges), style switching without stacking, and disable-after-custom-style. `tests/test_protocol.py`: message shape and validation for `style` / `width_px` / `threshold_deg`.

Screenshots in the next comment.


tdamsma:
Comment by Claude (bot)

Visual diff — sphere (smooth), capsule (curved + creases), box (prismatic), same scene and camera in every shot.

**Before — `edges` at THREE's 1° default (what `set_highlight` shipped in v0.0.47):** the reported X-ray wireframe.

before

**`edges` at the new 30° default:** the box keeps its 12 creases, but the sphere and capsule have no crease that sharp, so they get *no outline at all* — which is why the threshold knob alone doesn't close #165.

edges 30

**After — the new default `silhouette`:** a clean contour on all three, including the box (split normals and all).

after

**Disabled again:** original appearance restored exactly, no residue.

off

tdamsma:
Comment by Claude (bot)

Rebased onto `main`. This branch had been cut from a local `main` carrying three unpushed `append_points` commits, so the PR diff showed that feature alongside the highlight work. Those commits are now #174, and this branch is just the silhouette/edges change.

Two of the three Copilot notes were about the `append_points` code and have moved with it to #174 (the `ensurePointsCapacity()` bounds drop, and the `append_points` colors docstring). The one that still applies here is real: a silhouette hull relies on the mesh writing depth, but `set_opacity()` sets `depthWrite = false` below opacity 1, so a highlighted translucent mesh would render as a solid shell instead of a rim. The skinned/no-normals fallback to `edges` looks like the right place to also catch a non-depth-writing material.

tdamsma:
Comment by Claude (bot)

Fixed, and the root cause was a bit deeper than the review note.

Testing `depthWrite` alone would not have caught it. `_createMaterial` builds a primitive with `transparent: opacity < 1` but leaves `depthWrite` **true**, so `add_box(opacity=0.4)` still writes depth — and the shell appeared anyway. The reason is the render-pass split: three.js draws the whole opaque pass before the transparent pass, and `renderOrder` only sorts *within* a pass. The hull's material is opaque, so it always draws before a transparent mesh however high its renderOrder, and the depth buffer is still empty when it does.

So the predicate is `!transparent && depthWrite !== false` (`meshPrimesDepth`), folded into a new `resolveHighlightStyle` alongside the existing skinned-mesh and missing-normals fallbacks. The style is also re-resolved when `set_opacity` flips those flags under an already-highlighted mesh, so `set_highlight` → `set_opacity` and the reverse order land identically.

Measured before/after on a framed translucent box, counting selection-orange pixels against the object's footprint:

| | orange / covered |
|---|---|
| before | 394863 / 407362 (97%) — a solid shell |
| after | thin outline, asserted < 35% |

`test_set_highlight_composes_with_set_color_and_opacity` encoded the old behaviour (hull surviving `set_opacity`, material opacity 1.0) and now asserts the style flips to `edges` with the outline keeping its own 0.95 rather than the mesh's 0.5. Three new tests cover the fallback, the pixel-level shell regression, and the re-resolve in both directions plus reversibility after it. All 10 highlight tests pass.

**One consequence worth your call:** a *smooth* translucent body (a sphere) now falls back to `edges`, which has no crease above `threshold_deg` to draw — so it highlights as **nothing at all** rather than as a shell. Nothing is more honest than a shell that hides the object, but it is still a gap. Fixing it properly needs a depth-prime pass (draw the mesh's geometry `colorWrite: false, depthWrite: true` before the hull), which is doable but has its own ordering trap: the prime writes depth at the object's own surface, so the object's fragments then fail a LESS depth test unless offset. I did not attempt that here. Say the word if you want it as a follow-up.

PR #172, opened 2026-09-01
Support Draco-compressed GLB (KHR_draco_mesh_compression), decoder bundled — no CDN (#167)

### LINKED ISSUE #167: Support Draco-compressed GLB (KHR_draco_mesh_compression)
_Posted by Claude (bot) on behalf of Thijs._

## Motivation
The cell-setup-description visual mesh corpus is 198 MB of uncompressed GLB (largest: TRACK_XL_RAILS_18M_1.glb 37 MB; several extruders ~12 MB). Every /cells page load in ribweaver fetches the active cell's meshes over HTTP. Draco typically shrinks CAD-tessellated meshes 90–95%, so the corpus would drop to roughly 10–20 MB with visually identical results.

## What's needed
- Wire `DRACOLoader` into the existing `GLTFLoader` (`gltfLoader.setDRACOLoader(...)`).
- Ship the Draco WASM decoder files with the viewer bundle and point `setDecoderPath` at them — no CDN dependency (prod runs behind flexhub; offline dev must keep working).
- Uncompressed GLBs must keep loading unchanged (the extension is per-primitive; GLTFLoader falls back automatically).

## Out of scope here
Re-encoding the corpus (e.g. `gltf-transform draco` or `gltfpack`) is a cell-setup-description change and can land after the loader support ships — old viewers can't read Draco GLBs, so loader first, corpus second.

## Alternative worth a look
`EXT_meshopt_compression` (gltfpack + `MeshoptDecoder`) — smaller decoder, faster decode, slightly worse ratio. Either is fine; Draco has the broadest tooling.

tdamsma:
Closes #167.

## What shipped

`KHR_draco_mesh_compression` glTF/GLB now loads through the normal `add_model` / `add_model_binary` path. Both `GLTFLoader`s share one `DRACOLoader` (`viewer._dracoLoader`, one worker pool) attached with `setDRACOLoader`.

**The decoder ships inside the bundle — there is no CDN fallback at any level.** `build.py` gzips the vendored `static/draco/draco_decoder.wasm` + `draco_wasm_wrapper.js` (deterministically, `mtime=0`) and base64s them into a `DRACO_DECODER_DATA` constructor option, alongside the existing cubemap inlining.

`createDracoLoader(options)` resolves the decoder in precedence order:

1. `options.dracoDecoder` = `{wasmGzB64, wrapperGzB64}` — the `viewer.html` path, works under `file://`.
2. `options.dracoDecoderPath` — a URL prefix an embedder serves the two files from (three's `setDecoderPath` contract).
3. `static/draco/` resolved against `import.meta.url` — for an embedder serving the `viewer/` source dir as static files.

The inline path works by overriding **`DRACOLoader._loadLibrary(url, responseType)`**, the single place the class turns a decoder filename into bytes (normally a `FileLoader` GET under `decoderPath`). Everything else — worker pool, decoder config, `dispose` — is untouched, and `decoderPath` stays `''`, so no URL is ever constructed. It is a private method, so it is pinned to the vendored decoder pair and to `build.py`'s `THREE_VERSION`; a three upgrade that renames it fails loudly on the first Draco asset *and* in the browser test.

## Bundle size

| | bytes | delta |
|---|---|---|
| `viewer.html` before | 686,300 | |
| `viewer.html` after | 790,101 | **+103,801 (+15.1%)** |

Raw base64 of the same decoder pair would have been ~334 KB (+49%). The wasm is highly compressible (188 KB → 62 KB) and `DecompressionStream('gzip')` is native, so the inflate costs nothing worth measuring — and it only ever runs on a scene that actually contains a Draco primitive.

Vendored files are three r0.183.2's `examples/jsm/libs/draco/gltf/` pair (the glTF-targeted, decoder-only build: 188 KB wasm vs 279 KB for the general one). Added to `THIRD_PARTY_NOTICES.md`; Draco is Apache-2.0 and the license text was already in that file for the turbo colormap.

## Uncompressed GLB is unchanged

The extension is per-primitive and GLTFLoader falls back on its own. The decoder is also lazy — `DRACOLoader` inflates the wasm and spawns a worker only on the first compressed primitive — so an all-uncompressed scene never pays for any of this. Asserted in the test below (`workerPool.length === 0`).

## Tests

Two browser tests, both green, plus the full suite (476 passed; one unrelated pre-existing flake in `test_deferred_reparent_follows_parent_transforms` under full-suite load, green in isolation).

- `test_draco_compressed_glb_loads` — a Draco GLB decodes to the exact 4-vertex / 2-triangle unit quad, a worker was spun up, and `decoderPath === ''`. The page loads from `file://`, so this is a real offline exercise of the inlined path.
- `test_uncompressed_glb_still_loads_with_draco_wired` — a plain GLB loads with zero Draco workers.

**Fixture choice:** the compressed payload is a 75-byte Draco blob (POSITION only, EDGEBREAKER, 14-bit quantization) checked in as a base64 constant, with the GLB container built around it in Python mirroring the existing `_two_triangle_glb` — so the two fixtures differ *only* in the compression. It is not generated at test time because that would need a Draco *encoder*, which the project does not ship (I generated it once with three's `draco_encoder.js`; the exact recipe is in a comment above the constant so it is reproducible).

## Deliberately left out

- **Re-encoding the corpus** — out of scope per the issue; old viewers can't read Draco GLBs, so loader first, corpus second.
- **`EXT_meshopt_compression`** — the issue's alternative. Not wired; Draco has the broader tooling and doing both doubles the surface for no immediate gain.
- **No Python API surface.** Draco support is transparent — nothing in `client.py` changed, no new message type, no flag.
- **No `dispose()` hook for the DRACOLoader** — `ThreeJSViewer` has no `dispose()` method today, so there was nowhere to put it.
- The vendored decoder is excluded from `npx tsc` (`jsconfig.json`); it is minified emscripten output and type-checking it produces ~40 meaningless errors.


tdamsma:
Comment by Claude (bot)

Validated against the **real Draco assets** now on `cell-setup-description@adxc` (`glb-models/visual/draco/`), not just the synthetic fixture. All three are genuine `KHR_draco_mesh_compression` with the extension in `extensionsRequired`, so they fail hard without a DRACOLoader — a clean pass/fail for this PR.

Loaded each Draco GLB and its uncompressed twin side by side through `add_model_binary` on a `file://` page (i.e. the bundled, no-CDN decoder path):

| model | plain | draco | saving | meshes | triangles | max bbox delta |
|---|---|---|---|---|---|---|
| SUPPORT_BASE_1 | 103,628 B | 15,592 B | **−85.0%** | 8 = 8 | 3020 = 3020 | 0.3 mm |
| SUPPORT_EXTENSION_1 | 125,832 B | 18,648 B | **−85.2%** | 10 = 10 | 3892 = 3892 | 0.4 mm |
| SUPPORT_PIN_1 | 29,744 B | 3,368 B | **−88.7%** | 1 = 1 | 938 = 938 | 0.3 mm |

Mesh count and **index count are identical** in every pair, `decoderPath` stayed `''` (no URL ever constructed), 4 decode workers spun up, and the only console error was the harmless WS `426` from the test rig. Screenshot — Draco row and plain row are indistinguishable:

draco vs plain

Two notes, neither a viewer issue:

- **Vertex counts differ slightly** (3315→3324, 3958→4105, 964→970) while triangle counts match exactly. That is expected: the Draco encoder re-indexes and splits/merges vertices at attribute seams. Geometry is unchanged; only the index/vertex bookkeeping is.
- **Position quantization is ~0.3–0.4 mm** on these parts (bbox grows by ~0.6–0.8 mm overall, i.e. roughly 10-bit position quantization for a ~600 mm part). Invisible for a visual mesh, but worth knowing before anyone reuses a `visual/draco/` GLB as reference or collision geometry — it is an encoder setting on the producer side (`--quantize-position`), not something the loader can recover.

Issue #142, opened 2026-07-22
add_model_binary fetch ignores resp.ok → a 404/short blob body throws RangeError instead of a clean skip/retry


tdamsma:
> _Filed by an automated agent on Thijs's behalf (investigation of ribweaver board card C54)._

**Symptom.** When a `/api/blob/{key}` fetch 404s (or returns a short/empty body), the `add_model_binary` load path throws an uncaught `RangeError: Invalid typed array length: 4` (Chrome) / `attempting to construct out-of-bounds Uint8Array on ArrayBuffer` (Firefox), the mesh silently never renders, and there is no retry. Observed ~12× on a ribweaver `/cells` page reload for robot-link meshes (`cell_mesh_J1..J6`, `Base`, `EXTRUDER`, `pedestal_mesh`).

**Root cause.** `viewer.js` `case 'add_model_binary'` fetches and decodes without checking HTTP status:

```js
// viewer.js ~10210-10211
const resp = await fetch(data.blob_url);
const meshBytes = await resp.arrayBuffer();   // <-- no resp.ok check
const blob = new Blob([meshBytes]);           // 404 body → garbage GLB → RangeError
```

On a 404 the empty/error body is wrapped in a Blob and handed to the GLTF/STL loader, which reads a bogus chunk length and constructs an out-of-bounds typed array. The `add_object` direct-URL path (used for GLB fixtures) degrades more gracefully because three.js `FileLoader` surfaces the HTTP status as `…responded with 404`. Only the `add_model_binary` viewer-owned fetch lacks the guard. (`resp.ok` is checked in 2 other places in the file, not here.)

**Requested fix.** In `add_model_binary` (and any sibling binary-fetch cases), check `resp.ok` before `arrayBuffer()`; on non-2xx, reject/skip cleanly (optionally a bounded retry) rather than feeding the body to the loader. That converts a transient server-side blob miss into a recoverable no-op instead of an uncaught RangeError + dead mesh.

**Context.** The upstream trigger — ribweaver's shared 50-entry blob-store LRU evicting still-referenced cell meshes — is being fixed ribweaver-side (CEAD-group/ribweaver C54 issue). This viewer guard is the defense-in-depth half: it makes any future blob miss non-fatal. Seen against threejs-viewer v0.0.46.

PR #173, opened 2026-09-01
Quiet the blob delete-then-fetch race on object re-push (#157)

### LINKED ISSUE #157: Blob delete-then-fetch race on object re-push (harmless 404 + console noise)
*Filed by a Claude agent on Thijs's behalf.*

Found during ribweaver /cells smoke hardening: re-pushing an object (delete + re-add of the same id, e.g. ribweaver's `cell_rex_*` preview objects) races the viewer's blob fetch — the old blob URL is revoked/deleted while a fetch for it is still in flight, producing a harmless 404 and console noise on re-preview.

- Symptom: intermittent 404 + console error on object re-push; no visual defect (the re-added object loads fine).
- ribweaver-side: the cells smoke now carries a narrow ignore for `cell_rex_` ids only — we'd like to drop it once the viewer tolerates (or cancels) the in-flight fetch on delete.
- Likely related seam: the blob_url↔url handling; the new deferred re-parenting (#139) and diagnosable load errors (#151) made this much easier to see.

tdamsma:
Fixes #157.

## Diagnosis

Re-pushing an object (`delete` + `add_*` of the same id) races the viewer's blob fetch. The producer — ribweaver's shared blob-store LRU, or this repo's own Python for animation blobs (`client.py` drops every `/animation_*` key on the next `load_animation`) and octree-LOD node blobs (`_release_points_lod` on `delete`/`clear`) — drops the old blob *as* it sends the delete. Meanwhile the viewer still has a `fetch` open for it, which 404s.

The viewer already had staleness machinery (`_sceneGeneration`, `_animGeneration`, per-id load tokens), but **every one of those checks sits after the `await`** — i.e. after the fetch has *succeeded*. A fetch that fails goes straight to the case's `catch`, which since #142/#146 emits a fully-attributed `console.error`. So a fetch belonging to an object that was already superseded reported exactly like a fetch for a live object: harmless 404, loud error, no visual defect.

There is also a genuine cross-transport race: the 404 arrives over HTTP, the `delete`/re-add that explains it over the WebSocket. Cancelling on delete fixes the ordering where the delete lands first, but not the one where the 404 wins.

## Fix (viewer-only; no Python or wire change)

**(a) Cancellation.** Every blob fetch is registered per id in `_loadAborts` (`_trackFetch(id, {supersede})` / `_untrackFetch` / `_abortFetches`) and carries an `AbortSignal` — `fetchArrayBuffer(url, what, signal)`. `_deleteObject` aborts the id's fetches (covering `delete`, `clear` and any non-binary re-add), and every binary **add** case supersedes — aborts the previous fetch for that id — before registering its own. Read-side fetches (`update_polyline_colors`, `update_parametric_tube_colors`, `set_follow_path`, the octree-LOD node stream) register *without* superseding, so they can't cancel a concurrent add for the same id, but still cancel with the object. When the delete wins the race, the request is gone before the blob is — no 404 is issued at all, not even in Chrome's network log.

**(b) Classification.** All binary-load catches now go through `_reportLoadFailure(err, what, loud, ctx)`. If `_isLoadSuperseded` holds (an `AbortError`, or a bumped `_sceneGeneration` / `_animGeneration` / load token), it logs one quiet `Discarding superseded <what> fetch for '<id>'` in the style of the existing `Discarding stale …` lines. Otherwise it runs `loud`, which is the **unchanged** `console.error` text of #142/#146. Because the 404 can beat its own explanation, a **fetch-stage** failure that still looks live is re-checked after `SUPERSEDED_FETCH_GRACE_MS` (500 ms, test seam `window.__supersededFetchGraceMs`) before shouting; post-fetch (parse) failures report immediately, and the load token is consulted only for the fetch stage — installing an object bumps its own token via `_deleteObject`, which would otherwise excuse a real post-fetch bug.

Not chosen: (c) retaining blobs Python-side. This repo's object blobs are never dropped in the first place — the deletes that cause this live in the embedder (ribweaver) and in the two Python paths that *must* free memory (animation swaps, LOD release). A retention window there would leak and still wouldn't cover the embedder.

### How the #142 distinction is preserved

The quiet path is entered **only** when the viewer can show the load is dead: the fetch was aborted by a delete/supersede, or the scene/animation generation or the id's load token moved on. A 404 for an object nobody deleted matches none of those, so it still produces the same `Error creating mesh 'x' after N bytes from <url>: … HTTP 404 …` console.error — just up to 500 ms later. The existing `test_binary_fetch_404_logs_and_skips` (issue #142) passes unchanged and asserts exactly that.

## Test

New `test_repush_during_blob_fetch_is_quiet` drives `handleMessage` directly and covers **both** orderings deterministically:

- **A** — add (dead blob URL) → `delete_object` → re-add (good in-page `Blob` URL), all in one synchronous turn: the fetch is aborted mid-flight.
- **B** — add (dead blob URL), wait for the 404 to land, *then* delete + re-add, with the grace widened via the test seam so the re-check window is not timing luck.

Asserts both re-added objects load as meshes, no `console.error` mentions either id, no page errors, and that the quiet classification actually fired (`Discarding superseded mesh fetch for '…'`) rather than the failure silently vanishing.

Verified the test fails on the pre-fix build with precisely the reported symptom (`Error creating mesh 'race_a': … HTTP 404 …` for both orderings). Full suite: 476 passed. `tsc --noEmit` and `ruff` clean. `CLAUDE.md` updated (Communication Model → HTTP status guard).

### Note for the ribweaver side

Ordering A leaves no console output at all. Ordering B still shows Chrome's own `Failed to load resource: … 404` network line — that request genuinely happened, and no page-level code can suppress it — but no viewer error accompanies it. So the `cell_rex_` ignore can drop the viewer-error match; if the smoke also greps browser network lines, keep a narrow ignore for those.


tdamsma:
Comment by Claude (bot)

Both addressed.

**`ANIMATION_FETCH_KEY`.** Rather than validating ids against the sentinel, it is now a `Symbol`. `_loadAborts` is a `Map`, which takes any key type, and a Symbol is not equal to any string an id could ever be — so the collision is impossible by construction and no validation is needed on the id path to keep it that way. The Map and the three tracking helpers widen from `string` to `string|symbol`; `tsc --noEmit` is clean.

**The 4.5s test wait.** Replaced, though not with a single `wait_for_timeout`. The re-check always logs *something* when the grace expires — the superseded discard or a loud error — so the test now polls for whichever lands and stops there, which is both faster and more robust than any fixed sleep (it cannot silently pass by out-waiting a window that moved). The grace seam also drops from 3000ms to 1000ms, still 2× production, and the delete it waits for is dispatched synchronously. The test now runs in **~4s instead of ~9s**.

PR #174, opened 2026-09-01
Add append_points: grow a point cloud in place at O(new points)

tdamsma:
Adds live point streaming: growing an existing point cloud in place instead of re-sending it. No issue was filed for this one — it came in as direct work.

## What it does

`ViewerClient.append_points(id, positions, colors=None)` grows a flat cloud created by `add_points` **in place**, O(new points) on both sides. Only the new points cross the wire (`append_points_binary` header + blob) and the viewer writes them into spare capacity at the tail of the existing GPU buffers — `addUpdateRange`, so the upload is a `bufferSubData` of the appended slice alone, not a re-upload of the cloud.

The motivating producer is a laser tracker at 1 kHz flushed a few times a second and grown toward ~10M points over an hours-long run. Re-sending the whole cloud with `add_points` is O(total) per update and stalls the browser around 1e5 points at ~4 updates/s.

## Design points

- **Capacity doubles when it runs out** (floor 1024). `ensurePointsCapacity` moves the data into a *fresh* `BufferGeometry` and `dispose()`s the old one — a replaced-but-not-disposed attribute leaks its GL buffer.
- **Appends are serialized per id** (`_pointsAppendChain`): their blob fetches otherwise complete out of order, and buffer order is the `draw_range` prefix contract. A partial reveal is left where it is; a fully-revealed cloud grows its draw range with the append.
- **Bounds are expanded, not recomputed** — box exact, sphere = the box's circumsphere (a safe over-estimate). `computeBoundingSphere` would be O(total) per append.
- **The colormap range is frozen at `add_points` time.** A later chunk's scalar colours clamp at the ramp ends rather than rescaling — rescaling means re-colouring, and re-uploading, every point already there, which is the cost this method exists to avoid. Pass explicit `cmin`/`cmax` when the range is known up front.
- **Unknown ids raise, they never vanish.** Python tracks flat clouds per id (`_points_flat`, dropped on `delete`/`clear`/replace-by-LOD), so an append to an id it never created — or to an octree-LOD cloud, or one with `birth_times`/`removal_times` — raises `ValueError` instead of losing stream data. The viewer additionally warns per append for a cloud it does not have, which is the browser-reload-mid-stream case: nothing replays a point cloud on reconnect, so re-seed with `add_points`.
- **Append blobs are evicted past 256 chunks** — one blob per update forever would retain every point an hours-long stream ever sent.

## Tests

Protocol tests cover the header/payload (new points only), RGB and scalar colours through the add-time colormap range (a value past `cmax` clamps to the same colour as `cmax`), every `ValueError` path, and the bounded blob history. Browser tests extend the `add_points` coverage: three appends onto a 100-point cloud land 250 points with one geometric growth to the 1024 floor (colours grown alongside), the draw range follows, the last chunk sits at the tail (send order preserved), bounds expanded; an append for an id the viewer does not have warns instead of silently dropping.

`examples/30_live_points_stream.py` seeds a cloud with a pinned colour range and grows it ~2500 points at 4 Hz until interrupted.

---

*Provenance: these three commits had been sitting on a local `main` ahead of `origin/main`, and were picked up as base commits by #168 and #170 — which is why both of those PRs briefly showed this feature in their diff. Both have since been rebased flat onto `main` and now contain only their own change.*


tdamsma:
Comment by Claude (bot)

All three review findings fixed.

**Bounds across a capacity growth.** Confirmed and fixed — the box and sphere are now cloned onto the grown geometry. Worth noting the cost is not what it first looks like: growth is geometric, so the re-seeding was amortised *linear* overall, not quadratic. What it actually did was put a full O(count) re-scan on the single append that happens to trigger a doubling — a latency spike on exactly the multi-million-point live stream this path exists for, which is reason enough.

It is only observable in how much work it does, so the test counts `Box3.expandByPoint` across a growing append: **900** for a 900-point append onto a 150-point cloud (plus 8 from scene-bounds bookkeeping expanding a box's corners), against 150 + 900 without the fix.

**Append blobs on `clear()`.** Fixed, and extended to `delete(id)` for the same reason — blob keys are now tracked id-tagged, and `_drop_append_blobs()` evicts one cloud's or all of them. The cap only ever bounded a *running* stream; a cloud that stopped under it kept its chunks for the life of the process.

**Colors docstring.** Corrected to describe what is actually enforced: only per-point-vs-flat. Scalar and (M,3) RGB chunks are interchangeable once a cloud has per-point colours, since both become the same RGB bytes on the wire.

PR #179, opened 2026-09-07
EDL composer: SMAA anti-aliasing pass after OutputPass

tdamsma:
## Summary

- The direct (EDL-off) render gets hardware MSAA from `antialias: true`, but that only covers the default framebuffer. Every `EffectComposer` target is single-sampled, so switching on eye-dome lighting (`Shift+D`, `set_edl`, or the point-cloud auto-enable) silently dropped all geometry-edge anti-aliasing.
- A multisampled composer target is not an option: the EDL pass samples the composer's depth texture, which a multisampled target cannot expose. So this appends an `SMAAPass` as the **final** composer pass, after `OutputPass` (SMAA's luma edge detection expects display-referred, tone-mapped sRGB colour).
- The SMAA blend quad is set to `NoBlending`, the same convention as `OutputPass`, so transparent-background pixels are written as `(0,0,0,0)` instead of blended over the previous frame (no trails, background colour unchanged).
- Cost is a fixed ~1 ms at 1080p / ~4 ms at 4K, independent of scene size. Nothing changes for the EDL-off path.

`_materialBlend` is a private member of three r183's `SMAAPass` (the bundled typings still say `materialBlend`); the code resolves either and throws loudly on neither, and the browser test pins it.

## Test plan

- [x] `test_depth_cue_edl_preserves_background` extended: SMAA is the last pass and its blend material is `NoBlending`
- [x] All EDL / depth-cue browser tests pass
- [x] `tsc --noEmit`, ruff clean, `viewer.html` rebuilt
- [x] Before/after screenshots (posted below)
- [x] Full suite: 516 passed

https://claude.ai/code/session_01NuwDJbWyw1EmSCWbJxR1Q8


tdamsma:
Comment by Claude (bot)

Before/after with EDL on, headless Chromium (SwiftShader, DPR 1). Bottom row is a 3x nearest-neighbour zoom of the boxed region. Left is main, right is this branch: the box silhouette, the fat turbo polyline, the 1px native line and the cylinder rim all lose their staircase.

before/after

Note the second commit: with a fourth swapping pass the composer's buffer parity flipped every frame, so the EDL depth texture (attached to `renderTarget2` only) was sampled from the wrong target and every other frame rendered blank. `renderComposer()` now swaps back to `renderTarget2` after each render, and the browser test asserts that invariant.

PR #185, opened 2026-09-08
Billboards, per-vertex mesh alpha, and visible= on every add_*

tdamsma:
Daniel's `billboard` branch (`origin/billboard`, 2 commits by Daniel Meursing), reviewed and pushed here with fixes on top. Three features arrived together:

## 1. `add_billboard` — a plane that faces the camera

`add_billboard(id, width, height, color, opacity, axis=None, lit=False, ...)`. `axis=None` is a **full** billboard (copies the camera quaternion, parallel to the view plane); `axis=[x,y,z]` is **cylindrical** (local +Y pinned to that world axis, spin from the camera direction projected perpendicular to it), so an upright label stays upright. Orientation is computed in world space and converted back into the parent's frame, so a billboard under a rotating group faces the camera instead of tumbling with it; with the camera exactly on the lock axis the last pose is held. Re-oriented from `_animate` *after* `controls.update()`. Otherwise a perfectly ordinary tracked object — delete/clear/set_color/set_opacity/parent/framing/picking all behave as for any primitive. Demo: `examples/32_billboard.py`.

## 2. `add_mesh(colors=...)` accepts `(N, 4)` RGBA

Component count travels as `vertexColorComponents` and becomes the color attribute's `itemSize`. Any RGBA mesh is forced `transparent: true` / `depthWrite: false` regardless of material `opacity` — a vertex alpha below 1 is invisible in an opaque material. RGB meshes are byte-identical on the wire. Demo: `examples/33_mesh_vertex_alpha.py`.

## 3. `visible=` on every `add_*`

`visible=False` creates the object hidden, instead of add-then-`set_visible(False)` which renders one frame of it first. The viewer applies it through one choke point, `_applyInitialVisibility`.

Also: sidecar blob URLs now use `127.0.0.1` when `host == "localhost"`, since the browser may resolve `localhost` to `::1` while the `HTTPServer` is bound to IPv4 — every binary fetch then fails while the WebSocket is fine.

## Review fixes I applied on top (last commit)

- **Two leftover debug `console.log`s** in the `add_mesh_binary` case (one of them dumping the whole material object every mesh).
- **Visibility-baseline ordering bug.** `_deleteObject` prunes the id's `_baselineVisibility` entry, so `_applyInitialVisibility` has to run *before* it. The six binary add paths called it after, which silently dropped a `set_scene_visibility` that arrived mid-fetch — precisely the race that map exists for. Non-binary paths were already correct.
- **CLAUDE.md**: the branch documented the billboard and example 33 only; added the `32_billboard.py` entry and the three undocumented behaviours (the `visible=` kwarg + its baseline interaction, RGBA vertex colors, the sidecar host).

## Verified

`ruff check/format`, `tsc --noEmit`, and the full suite locally: **360 unit + 179 browser tests pass**, including Daniel's new browser tests for the RGBA attribute and for `visible=False`.

https://claude.ai/code/session_01QoVmF37jo29rBEbqNv8V81


tdamsma:
Comment by Claude (bot)

Both Copilot comments were the same finding (source + generated bundle), and it was right: fixed in 9ee1a17 (or whatever the latest commit is).

`buildBillboardMesh` now validates `axis` — finite, non-zero, length 3 — and degrades to a full billboard with one `console.warn` otherwise. Python already rejected those inputs, but `handleMessage` is a documented public no-WS embedder surface, so the viewer has to hold the line itself.

One correction to Copilot's framing: a zero axis does *not* reliably produce NaN. `Vector3.normalize` divides by `this.length() || 1`, so a zero vector survives as a zero vector; what actually happens is that `makeBasis` gets zero columns and `setFromRotationMatrix` reads a degenerate basis, every frame. Corrupt either way, just not necessarily NaN — the code comment and test name say that rather than overclaiming.

Also added two browser tests, since the branch had none for the billboard itself:
- the orientation contract — a full billboard's world quaternion equals the camera's *including under a rotated parent* (the case the world→parent conversion exists for), and a Z-locked one keeps local +Y on world +Z;
- both degenerate axes driven straight through `handleMessage`.

Regenerated `viewer.html` from source, as the comment on the bundle asked.

PR #186, opened 2026-09-08
Orient-to-viewer as a per-object property (set_billboard)

tdamsma:
Follows #185. Plan approved in-session.

`add_billboard` could only face a *plane* at the camera, and did it by **overwriting** the object's rotation using a **world-space** axis. Neither is expressible in terms of the model you are actually looking at, and it ruled out the obvious case: **a wheel that keeps spinning about its axle while the axle points at the camera.**

## The model

Facing is now a property any object can carry — primitive, mesh, model or group — and every mode is one composition, applied per frame:

```
R_final = R_aim · R_own
```

`R_own` (the object's own rotation, authored or animated) acts **first, in the object's local frame**. That single choice is what makes the wheel work: a spin about an axis leaves that axis fixed, so the spin never fights the aim. The modes differ only in how many DOF the viewer claims:

| `mode` | DOF | Own rotation | Meaning |
|---|---|---|---|
| `camera` (default) | 3 | overridden | copies the camera's orientation — parallel to the view plane, no swing off-axis. Today's `axis=None`. |
| `world_up` | 3 | overridden | aims `face` at the camera, roll from `world_up`. |
| `aim` | 2 | **kept as roll** | minimal rotation putting `face` on the camera. **The wheel.** |
| `hinge` | 1 | kept | spins about `hinge` only. Today's cylindrical billboard. |

`face`, `up`, `hinge`, `pivot` are read in the **object's own frame** (axis names or vectors). `hinge_world` is the one deliberately world-space knob — a hinge needs both *which local axis is the hinge* and *which way it points in the world*, which the old `axis=` conflated. `pivot` is the rotation origin: a local point held fixed while the object re-orients, defaulting to the origin (position untouched).

## Bug fix from #185

An object driven by a `transforms` channel, a JSON `Frame.transforms`, or a follow path has `matrixAutoUpdate = false`, so `updateMatrix()` never runs and the old `obj.quaternion` write was **dead**. An animated billboard did not billboard at all. Confirmed by reverting the write path: the wheel's axle reads `[0,1,0]` instead of `[1,0,0]` — not aimed. Both new animated tests fail against the old behaviour and pass now.

## The load-bearing detail: a retained base pose

The composed result must never be read back as `R_own`, or each frame composes on its own output. This is not merely tidier — `_updateBillboards` runs every frame, but `_applyFrame` runs **only while an animation is playing** (`_animate` gates it). Pause, then orbit — the normal interaction — and nothing refreshes `obj.matrix`, so a read-back implementation integrates a fresh `R_aim` every frame into unbounded runaway. Every transform writer therefore records the producer's pose (`_applyTransform`, both checked matrix setters, `_applyFollowPaths`, gizmo drag-release).

Also handled: billboards iterate **parents-first** (insertion order leaves a nested billboard a frame behind — angular alone, but *positional* once a pivot is in play); a degenerate frame re-applies the last good aim rather than skipping the write; a gizmo drag owns the transform for its duration and its final pose becomes the new base.

## Demo — checkered donuts in every configuration

`examples/34_billboard_modes.py`. A torus is rotationally symmetric, so the checker is what makes roll and spin visible. Modes row / spinning wheel / face axis / pivot / hinge. The donut mesh is built in the example with `add_mesh` (one quad per check, unshared vertices for crisp edges, analytic normals) — no new primitive API.

## Breaking change (unreleased)

`add_billboard(axis=[0,0,1])` (world) → `mode="hinge", hinge="+y", hinge_world=[0,0,1]` (local hinge + its world direction). #185 is not in a release (last is 0.0.50), so example 32 and its tests are migrated in place rather than carrying two meanings of `axis`.

## Verified

`ruff`, `tsc --noEmit`, regenerated `viewer.html`, **573 tests pass** (unit + browser). Nine new browser tests, each pinning a hazard from the design review: no-drift over 30 still frames, paused-animation + orbit, the spinning wheel, pivot anchor invariance, exact restore on `enabled=False`, nested no-lag, hinge axis held, degenerate `face ∥ hinge`. Screenshots in the first comment.

https://claude.ai/code/session_01QoVmF37jo29rBEbqNv8V81


tdamsma:
Comment by Claude (bot)

### Checkered donuts, every configuration

Rows top to bottom: **modes** (camera / world_up / aim / hinge) · **wheel** (aim + its own spin) · **face axis** (+z / +y / +x) · **pivot** (centre / rim) · **hinge** (hinge_world None / +Z).

modes overview

The modes row is spread wide on purpose: `camera` (leftmost) stays parallel to the screen, `aim` (third) visibly toes in toward the camera, and `world_up` (second) differs from `camera` only in roll — which you can read off the checker phase, and could not see at all on a plain torus. `hinge` (rightmost) is the 1-DOF case, tilted because it may only spin about its upright hinge.

front view

### The wheel: aim + its own spin

Same camera, ~1 s apart. The axle stays pointed at the viewer while the donut keeps turning about it — `R_final = R_aim · R_own` with the spin arriving on a binary `transforms` channel.

| t = 2.10 s | t = 3.59 s |
|---|---|
| spin a | spin b |

Before this PR that path produced **no** billboarding at all: the channel pins `matrixAutoUpdate = false`, so the old `obj.quaternion` write was silently dropped. I confirmed it by reverting just the write branch — the axle reads `[0,1,0]` instead of `[1,0,0]`, and both animated tests fail.

PR #195, opened 2026-09-15
Apply data.transform in the polyline and point-cloud add handlers

### LINKED ISSUE #194: add_polyline_binary ignores data.transform
> Posted by Claude Code on Thijs's behalf.

The `add_polyline_binary` handler (`viewer.js:11727-11868`) never reads `data.transform`, while `add_group` (`:11414`), `add_model` (`:11706`) and the mesh handlers (`:12302`, `:12624`, `:12784`) all apply it. A polyline sent with `transform.matrix` therefore lands at the world origin.

Found via ribweaver's cell workzone wireframe, which is the one extras polyline carrying its own matrix instead of sitting under a transformed group (ribweaver issue linked below). Ribweaver will work around it by parenting to a group, so this is about protocol consistency: either apply `transform` in `add_polyline_binary` like the other add_* handlers, or document that polylines never carry one.

https://claude.ai/code/session_011V7JUKwVevS1Agb69siXrD

tdamsma:
PR by Claude (bot) on Thijs's behalf.

Closes #194

## Viewer handlers

Audited every `add_*` case in `handleMessage` for a missing `data.transform`. Three handlers ignored it; each now calls `_applyTransform` right after `_registerObject`, the same ordering `add_mesh_binary`, `add_parametric_tube_binary` and `add_swept_tool_binary` use (visibility first, then delete/attach/register, then the pose, so a pending re-parent and the billboard base capture see the final matrix):

- `add_polyline_binary` (the issue; covers fat `Line2`, native `THREE.Line` and `segments=True` `LineSegments`, which share one registration tail)
- `add_points_binary` (flat point clouds)
- `add_points_lod` (octree group; `_updatePointsLOD` already inverts `group.matrixWorld`, so refinement is correct under a transformed group)

Already correct, left untouched: `add_group`, `add_object`, `add_model`, `add_model_binary`, `add_mesh_binary`, `add_parametric_tube_binary`, `add_swept_tool_binary`, `add_grid`, `add_billboard`. `register_toolpath_group` annotates a group created by `add_group`, which applies the transform itself.

## Python

`add_polyline` and `add_points` (both the flat and the LOD path) now accept `position`/`rotation`/`scale`/`matrix` like `add_mesh` and send them as the header `transform`. The header construction that was inlined four times (`add_model_binary`, `add_mesh`, `add_parametric_tube`, `add_swept_tool`) moved into one `_transform_header` helper; the wire format is unchanged and an untransformed add still omits the key.

## Tests

- `tests/test_protocol.py`: message shape for polyline (none / position+rotation / matrix wins / segments) and points (none / position+scale / LOD matrix).
- `tests/test_browser.py`: `getObject(id).position` and `matrixWorld` reflect the translation for all three polyline variants, a matrix under a translated parent group, a flat point cloud, and an octree-LOD cloud. All five browser cases fail against the previous `viewer.html` and pass with the rebuilt one.

`viewer.html` regenerated with `build.py`. One sentence added to the Polylines entry in CLAUDE.md.

Full run: `uv run pytest -q` gives 586 passed, 1 failed (`test_set_view_cancels_drag_inertia`, a camera-drag timing test unrelated to this change; it passes when rerun alone).


tdamsma:
Comment by Claude (bot) on Thijs's behalf.

Addressed the Copilot finding on the LOD node-size estimate. `_updateOnePointsLOD` took `setFromMatrixScale(group.matrixWorld).x` as the group scale and measured camera distance in cloud-local space, so a non-uniform scale such as `[1, 1, 100]` (now reachable through `add_points(scale=...)`) under-estimated node screen size on the stretched axis and stopped refinement early. The estimate now runs in world space: the node radius is multiplied by the largest of the three scale components (a conservative bound on the scaled cube's circumsphere) and the perspective distance is taken from the camera to the node center transformed by `group.matrixWorld`. Uniformly scaled and unscaled clouds produce the same numbers as before.

New browser test `test_points_lod_nonuniform_scale_uses_max_component`: three identical clouds at scale `[1,1,1]`, `[1,1,4]` and `[4,4,4]` under an orthographic camera (where the estimate is position-independent). With `refine_pixels=100` the unscaled cloud wants only the root node while the `[1,1,4]` and `[4,4,4]` clouds want the identical full set of 73 nodes. Under the previous code the `[1,1,4]` cloud matched the unscaled one.

`viewer.html` rebuilt; ruff clean; full suite 587 passed with one unrelated fixture-setup error in `test_parametric_tube_frontier_morph_positions` that passes when rerun alone.

PR #201, opened 2026-09-15
Bind the WS server and blob sidecar on every address host resolves to

### LINKED ISSUE #187: Sidecar blob URLs: localhost vs 127.0.0.1 (reverted #185's rewrite — what did it fix?)
_Filed by Claude Code on Thijs's behalf._

## What happened

#185 added `self._http_host = "127.0.0.1" if host == "localhost" else host`, so blob URLs advertised to the browser used `127.0.0.1` while the WebSocket URL kept `localhost`.

That makes **Firefox refuse every sidecar fetch from a `file://` viewer page**, so nothing loads at all — no meshes, no animation, an empty scene with no obvious cause. Console:

```
GET http://localhost:5666/                       426 Upgrade Required   <- reached the server
http://127.0.0.1:5667/blob_9f17a6be...  Cross-Origin Request Blocked:
    (Reason: CORS request did not succeed). Status code: (null)
Error creating mesh 'mode_camera': TypeError: NetworkError when attempting to fetch resource.
Error loading animation via HTTP: TypeError: NetworkError ...
```

Same page, same machine, same moment: `localhost` reaches the server, `127.0.0.1` is blocked. Reverted in #186 (`_http_host = host`).

**Chromium allows it**, which is why the browser suite passed throughout — the tests are Chromium-only, so this whole class of bug is invisible to CI.

## The open question (@DAlexanderr)

The rewrite was presumably fixing something real. The usual suspect is IPv6: a browser resolves `localhost` to `::1` while `HTTPServer` is bound to IPv4 only, so every blob fetch fails with a connection error even though the WebSocket (which the browser reaches by its own URL) is fine.

If that is what you hit, **the revert re-breaks you**, and the fix belongs at the *bind*, not in the advertised URL:

- bind the sidecar dual-stack (`AF_INET6` with `IPV6_V6ONLY` off) so `localhost` works whichever way it resolves, and keep advertising `host` verbatim; or
- bind to the address `localhost` actually resolves to on this machine (`socket.getaddrinfo`), still advertising `host`.

Either keeps one hostname in the page, which is what Firefox needs.

Could you say what the symptom was — OS, browser, and whether the viewer page was `file://` or served over HTTP? That decides which fix to build.

## Worth doing regardless

The browser suite runs Chromium only, so a same-origin/local-network policy difference like this cannot be caught. A single smoke test under Firefox (`pytest-playwright` supports `--browser firefox`) that loads one binary asset from a `file://` page would have caught it in seconds.

tdamsma:
PR by Claude (bot) on Thijs's behalf.

Closes #187

## Problem

#185 rewrote the advertised blob URL host from `localhost` to `127.0.0.1`, which made Firefox refuse every sidecar fetch from a `file://` viewer page as cross-origin (empty scene, "CORS request did not succeed"). #186 reverted it. The reason for the rewrite was IPv6: the browser can resolve `localhost` to `::1` while `HTTPServer(("localhost", port))` binds only `127.0.0.1`, so every blob fetch fails with a connection error on such machines. Both symptoms come from the same place, and the fix belongs at the bind, not in the advertised URL.

## Approach: one listener per resolved address

`_listen_sockets(host, port)` binds one listening socket per `(family, address)` that `socket.getaddrinfo(host, port, SOCK_STREAM, AI_PASSIVE)` returns (for `localhost`: `127.0.0.1` and `::1`). `_start_servers` runs one `_BlobServer` (an `HTTPServer` over a pre-bound socket) and one `websockets` sync `Server` (`serve(sock=...)`) per socket, all sharing the blob store and the connection handler. Blob URLs, the WS server, and the viewer URL all advertise `host` verbatim; `_http_host` is gone.

Why not the dual-stack `AF_INET6` + `IPV6_V6ONLY=0` server suggested in the issue: dual-stack only works on the unspecified address `::`. Binding `::1` with V6ONLY off still accepts IPv6 loopback only, and binding `::` would expose a `localhost` client on every interface. Per-address listeners keep `localhost` loopback-only on both families.

Details:
- IPv6 sockets are V6ONLY (what `socket.create_server` does without `dualstack_ipv6`), so the families never collide on a wildcard host.
- An IPv6 address that fails to bind is skipped with a warning (IPv6 disabled in the kernel or a container); at least one socket must bind or the error is raised.
- `http_port=0` (tests, bench scripts) picks a free port on the first socket and reuses it on the others, retrying on `EADDRINUSE`.
- A bind failure in `_start_servers` runs `disconnect()` so a half-started client does not leak listeners.
- The `websockets` sync server was IPv4-only too: `serve(handler, host, port)` calls `socket.create_server((host, port))` once with the default `AF_INET` family (verified in the installed `websockets==16.0` source), so it needed the same treatment. Both sides now bind synchronously in the caller, so bind errors surface in `connect()`.
- `disconnect()` drains `_http_servers` / `_ws_servers`; the sidecar `serve_forever` poll interval is 0.05 s so shutting down several servers stays fast. `test_timeout_cleans_up_servers` asserts on the new lists.
- `conftest.viewer_client`, `test_browser._start_client`, and the two `bench/tube_lab` scripts now call `client._start_servers(http_port=0)` instead of hand-building an `HTTPServer` and a `_run_server` thread.

## Tests

- `test_servers_reachable_on_both_loopback_families[127.0.0.1 | [::1]]`: fetches a blob over each family and TCP-connects to the WS port over each; the IPv6 case skips when `::1` cannot be bound.
- `test_blob_url_host_is_the_configured_host`: blob URLs carry `host` verbatim for `localhost`, `127.0.0.1`, and a custom name.
- `test_firefox_file_page_loads_binary_asset`: launches `playwright.firefox`, loads the `file://` viewer page, `add_mesh`, asserts the mesh lands in `query_scene`. Skips with `pytest.skip` when the Firefox binary is not installed. CI installs `chromium firefox`.

## Verified on macOS (Darwin 24.6)

- `localhost` resolves to `127.0.0.1` and `::1`; the pre-fix `HTTPServer(("localhost", 0))` bound `127.0.0.1` only and a connect to `::1` was refused. With this change a blob fetch over `http://[::1]:<port>` succeeds.
- Firefox smoke test passes (Playwright Firefox 146).
- `uv run pytest -q`: 578 passed (Chromium suite + Firefox smoke test). `ruff check` and `ruff format --check` clean.

## Not verified

Playwright's Firefox build does not reproduce the #185 refusal: with the blob host forced to `127.0.0.1` while the page uses `localhost`, the mesh still loads there, under the default prefs and with `security.fileuri.strict_origin_policy` / `network.lna.*` toggled. So the smoke test guards the file://-page-plus-sidecar path in Firefox but would not have failed on #185 as written. The refusal reported in #187 came from a stock Firefox; driving the installed desktop Firefox from the test run was not attempted beyond one probe that did not connect.


tdamsma:
Comment by Claude (bot) on Thijs's behalf.

Two follow-up commits on this branch.

**Firefox CI failure.** The first CI run failed the Firefox smoke test with "did not connect to the WebSocket server". Capturing the page console on the runner showed the real cause: headless Firefox on the GPU-less Ubuntu runner cannot create a WebGL context (`THREE.WebGLRenderer: Error creating WebGL context`), so the viewer constructor throws before `connect()` runs. Networking is not involved; locally (macOS, Playwright Firefox 146) the test passes, and Firefox connects fine even when the servers listen on one loopback family only. The test now launches Firefox with software-WebGL prefs (`webgl.force-enabled`, `webgl.forbid-software=false`, `gfx.webrender.software`), probes for a WebGL2 context on a blank page, and skips with "Playwright Firefox cannot create a WebGL2 context here" when there is none; the connect failure message includes the page console. On the runner it skips; on a machine with WebGL it runs for real.

**Copilot findings.**
1. `_listen_sockets` skips an IPv6 bind only for `EAFNOSUPPORT` / `EADDRNOTAVAIL` / `EPROTONOSUPPORT` (`_IPV6_UNAVAILABLE_ERRNOS`). `EADDRINUSE` on a requested port raises; with `port=0` it goes through the retry path. New test: a port taken on `::1` makes `_listen_sockets("localhost", port)` raise and leaves nothing bound.
2. `tools/capture_strand_collapse.py` uses `_start_servers(http_port=0)`; the manual HTTPServer setup and unused imports are gone.
3. `_start_servers` keeps the bound-but-not-yet-served sockets in a `pending` list and closes them on the exception path before `disconnect()`.
4. A blob server is appended to `_http_servers` only after its thread started, so `disconnect()` never calls `HTTPServer.shutdown()` on a server that will not enter `serve_forever`.
5. Pass-through: a non-`localhost` `host` is added to `viewer_url` as `ws_host`, and `viewer.js` builds the WebSocket URL as `ws://${ws_host || 'localhost'}:${ws_port}` (`viewer.html` rebuilt). `_url_host` brackets IPv6 literals in blob URLs and the param. Tests: unit test on `viewer_url` (absent for localhost, `127.0.0.1`, `[::1]`), and a Chromium test that `ViewerClient(host="127.0.0.1")` connects the WebSocket to `ws://127.0.0.1:<port>` and loads a mesh from that host.
6. The loopback test opens a real WebSocket over `127.0.0.1` and `[::1]` with the `websockets` sync client and asserts the server-side connection handler ran.

Local: ruff clean, `uv run pytest -q` 584 passed.

PR #204, opened 2026-09-15
Toolbar as a hidden-by-default menu listing the advanced options with their shortcuts

tdamsma:
## Summary

- The row of toolbar buttons becomes a single `☰` menu button at top-left (the connection-status dot rides on it). Its dropdown lists every advanced option with its keyboard shortcut and current state: Clipping plane `C`, Lighting `E`, Orbit mode `R`, Camera tracking `T` (only with a track target, as before), Projection `O`, Wireframe `M`, Shading debug `N`, Distance fog `D`, Eye-dome lighting `Shift+D`, Frame all `F`. Items run the same action as their key; a click outside or `Escape` closes the menu.
- **The menu button is hidden by default.** A bare viewer opens with no chrome besides the view gimbal. Keyboard shortcuts work either way.
- Enabling it: URL `toolbar=true` > `ThreeJSViewer({toolbar: true})` > hidden; `viewer.setToolbarVisible(bool)` / `isToolbarVisible()` at runtime. From Python: `ViewerClient(toolbar=True)` rides the viewer URL (survives a refresh on its own), and `set_toolbar_visible(bool)` sends a `set_toolbar` message that is replayed on reconnect.
- The `_btnClip`/`_btnLighting`/`_btnOrbitMode`/`_btnTrack` members now point at menu items, so the existing `.active`/`title`/`display` writers are untouched. `_updateOrbitModeButton` writes the state span instead of overwriting the button text.

## Screenshots

(added in a comment below)

## Test plan

- [x] `uv run pytest` (unit) and the full Playwright suite: 164 browser tests pass
- [x] New tests: hidden by default + `set_toolbar_visible` round trip, menu item list with shortcut keys and state labels, click-outside close, `toolbar` constructor option
- [x] `npx tsc --noEmit -p jsconfig.json`, ruff check/format


tdamsma:
Comment by Claude (bot)

**Default (hidden):** nothing but the gimbal.

default hidden

**`ViewerClient(toolbar=True)` / `set_toolbar_visible(True)`:** one menu button top-left with the connection dot.

menu closed

**Menu open** (after pressing `M` once, so the Wireframe item shows its active mark and `wire` state):

menu open

PR #208, opened 2026-09-17
Primitive material flags: wireframe, depthWrite, polygonOffset, and double-sided translucency

### LINKED ISSUE #207: add_object drops params.wireframe and params.depthWrite; applyOpacity clears depthWrite one-way
_Filed by Claude (agent) on behalf of @tdamsma, from ribweaver `/cells` work. Line references are against `2ca924d` (Release 0.0.52)._

## Summary

`add_object`'s primitive path silently drops two material flags that callers already send: `wireframe` and `depthWrite`. Separately, `applyOpacity` mutates `depthWrite` in a way that `_createMaterial` does not, so the same visual state reached two different ways renders differently, and the change is one-way.

ribweaver has been sending `wireframe: true` on every `/cells` collision box for as long as that feature has existed, and they have been rendering solid the whole time. Nobody noticed because a translucent solid box looks plausible on its own — it only became obvious when we raised the opacity.

## 1. `params.wireframe` is never read

`_createMaterial` (viewer.js:10012) reads exactly `color`, `materialType`, `opacity`, `roughness`, `metalness`:

```js
_createMaterial(params) {
    const color = params.color || 0x4a90d9;
    const materialType = params.materialType || 'standard';
    const opacity = params.opacity != null ? params.opacity : 1;
    const transparent = opacity < 1;
    // ...
    return new THREE.MeshStandardMaterial({ color, roughness, metalness, opacity, transparent, clippingPlanes: clip });
}
```

`params.wireframe` appears nowhere on the add path — the only `mat.wireframe` writes in the file are the M-key debug cycle at viewer.js:5946.

Concrete caller (ribweaver `web/shared/scene_manager.py`, one per collision box):

```json
{"type": "add_object", "id": "col_J4_0", "parent": "col_grp_J4",
 "object": {"primitive": "box",
            "params": {"width": 0.22, "height": 0.22, "depth": 1.4,
                       "color": 52479, "opacity": 0.25,
                       "transparent": true, "wireframe": true},
            "visible": true}}
```

Expected: a cyan wireframe cage around the link, so the operator can see the arm's geometry through it.
Actual: a solid 25%-opaque cyan box.

Why it matters to us: we now recolour a box red when the robot is in collision at the current pose, and the whole point is to see *how* the two bodies touch. A solid box hides the contact and everything behind it. We had to pick the highlight opacity assuming a solid box (0.45) rather than the value that would read well on a cage.

Note `transparent` is dropped too, though that one is harmless — `_createMaterial` derives it from `opacity < 1`.

## 2. `params.depthWrite` cannot be set

Same function, same cause. There is no way for a caller to ask for a translucent overlay that does not write depth, which is the normal way to stop coplanar or nested overlays fighting each other.

Our case: `/cells` stacks reach-analysis layers over a work surface. We currently separate them with literal millimetres of Z offset, which is a workaround for not being able to ask for `polygonOffset` or `depthWrite: false`.

## 3. `applyOpacity` and `_createMaterial` disagree about `depthWrite`, one-way

`applyOpacity` (viewer.js:1220):

```js
mat.transparent = opacity < 1;
mat.opacity = opacity;
mat.depthWrite = opacity >= 1;
```

`_createMaterial` never sets `depthWrite`, so a primitive created with `opacity: 0.25` gets three.js's default `depthWrite: true`. Reaching the *same* opacity through `set_color`/`set_opacity` gets `depthWrite: false`.

Repro, entirely through documented messages:

1. `add_object` a primitive with `params.opacity = 0.25` → `transparent: true`, `depthWrite: true`.
2. `set_color` with `{"color": 0xFF2020, "opacity": 0.45}` → `depthWrite: false`.
3. `set_color` back to `{"color": 0x00CCFF, "opacity": 0.25}` → `depthWrite` stays `false`.

Step 3 is the problem: the object is visually back where it started but sorts and blends differently from an identical neighbour that was never recoloured, and nothing short of re-adding it restores the original state. In our scene that means collision boxes which have been red at some point stop occluding each other, while ones that never collided still do.

This one is already half-documented in your own source — the `meshPrimesDepth` comment at viewer.js:1355-1363 notes that "`_createMaterial` sets `transparent` but leaves `depthWrite` true", which is exactly the asymmetry.

## What we would like

1. `_createMaterial` honours `params.wireframe` and `params.depthWrite` on the `add_object` primitive path, so the payloads we already send start meaning what they say.
2. Ideally `params.polygonOffset` / `polygonOffsetFactor` / `polygonOffsetUnits` too — that is the correct fix for coplanar overlays and would let us drop the millimetre Z offsets.
3. A decision on (3): either `_createMaterial` should apply the same `depthWrite = opacity >= 1` rule so the two paths agree, or `applyOpacity` should leave an explicitly-set `depthWrite` alone. Either is fine for us; the current split is what bites.

Happy to test a branch — ribweaver's `/cells` lane exercises all three paths.

tdamsma:
Closes #207.

`_createMaterial` read only `color`/`materialType`/`opacity`/`roughness`/`metalness`, so `params.wireframe`, `params.depthWrite` and `polygonOffset` were dropped, and it never set `depthWrite` at all while `applyOpacity` sets it to `opacity >= 1`.

**Now:**

- `wireframe`, `depthWrite` and `polygonOffset`/`Factor`/`Units` are honoured on the primitive add path.
- `depthWrite` defaults to `opacity >= 1` on both paths, so a primitive added translucent sorts like one turned translucent later. An explicit `depth_write=` outranks the rule and `applyOpacity` leaves it alone (`userData.__depthWriteExplicit`).
- Translucent primitives are double-sided, so a transparent body shows its far walls instead of a flat one-shade silhouette. Opaque bodies and wireframe cages stay front-only; `side="front"|"back"|"double"` overrides.
- Python: `wireframe=`, `side=`, `depth_write=`, `polygon_offset=`, `polygon_offset_units=` on box/sphere/cylinder/capsule. Payloads that set none of them are unchanged on the wire.

Note: a wireframe or `depth_write=False` primitive no longer primes the depth buffer, so `set_highlight` uses the `edges` style on it, as it already did for a translucent one.

## Before / after

Both scenes are the same script run against `main` and against this branch.

**Cages** — ribweaver's `/cells` payload (`opacity: 0.25, wireframe: true`), middle box recoloured red at 0.45 and back to cyan at 0.25.

| main | this branch |
|---|---|
| before | after |

The flag is ignored on `main`, so the cages render solid. `col0` and `col1` also end up with different `depthWrite` there (`true` vs `false`) despite identical opacity; on this branch both are `false`.

**Translucent solids** — same primitives without the wireframe flag: two overlapping boxes, a rotated one, and a sphere around an opaque core.

| main | this branch |
|---|---|
| before | after |

On `main` the boxes write depth, so they hide the grid and each other, and front-only culling leaves each one a flat silhouette. On this branch they blend, and the far walls and interior edges are drawn.

## Tests

5 protocol + 4 browser tests: the payload shape, `wireframe` and `polygonOffset` reaching the material, the `add` → `set_color` → restore → opaque `depthWrite` round trip, an explicit `depth_write` surviving `set_opacity`, and the double-sided default with its three exceptions. Full suite: 661 passed.


tdamsma:
> Posted by Claude Code on Thijs's behalf.

Two notes from the ribweaver side, verified against this branch (full suite: 661 passed).

**Workaround to remove when the pin bumps.** ribweaver carries a workaround for #207 in `ribweaver/web/cells/reach.py` (millimetre Z offsets separating the reach-analysis layers, standing in for `polygonOffset`/`depthWrite: false`) and `ribweaver/web/shared/scene_manager.py` (collision-box highlight opacity 0.45, chosen for a solid box because `wireframe: true` was being dropped). Both come out once the pin moves past this fix, and the highlight opacity should be re-tuned for a cage.

**The `DoubleSide` default deserves its own judgement.** Defaulting `side` to `DoubleSide` for translucent non-wireframe primitives is outside what #207 asks for, and it changes how every existing translucent primitive renders for every caller, not just the ones that opted in. The reasoning and screenshots are convincing on their own, but it is a separate behaviour change riding along with the flag fix and worth reviewing as such.

PR #209, opened 2026-09-17
Outline accents for translucent primitives, plus mesh transparent/renderOrder

tdamsma:
Outline work by @danielmeursing (branch `outline-geometry`), with Python senders and tests added. **Based on #208**, which the branch overlapped in `_createMaterial`; retarget to `main` once that merges.

**What this adds:**

- `outline=True` on `add_box`/`add_sphere`: a crisper line accent over the fill, `outline_opacity=` its opacity (default 0.9). A box gets its feature edges (`LineSegments`), a sphere a ring (`LineLoop`) kept camera-facing by the billboard system (`mode: 'aim'` — the ring's +Z normal is already that system's default `face` axis). An outlined primitive is a `Group`; un-outlined ones stay a single `Mesh`.
- `transparent=` and `render_order=` on `add_mesh`: three finishes the opaque pass before the transparent one and `renderOrder` only sorts within a pass, so a coplanar layer stack that wants paint order needs both.
- Outline children carry `userData.__primitiveOutline`: `set_color` recolours them, `applyOpacity` skips them (an accent that follows the fill's opacity dissolves), `set_visibility` hides them, and the `M`/`N` debug cycles skip them since `_forEachUserMesh` gates on `isMesh`. The sphere ring is registered as a billboard under `<id>::outline` but not in `_objects`; `_deleteObject` still prunes it.

All options are omitted from the payload when unset, so existing messages are unchanged on the wire.

## Before / after

Two overlapping translucent boxes, a rotated one, and a translucent sphere around an opaque core.

| `outline` off | `outline=True` |
|---|---|
| off | on |

## Tests

4 protocol + 4 browser: outline child types, the ring's billboard registration, the `set_color`/`set_opacity` composition, clean deletion, and the mesh render flags. Full suite: 669 passed.


tdamsma:
Comment by Claude (bot)

Superseded by #210 — GitHub closed this automatically when `primitive-material-flags` was deleted on the merge of #208. Same branch, same content, retargeted to `main`.

